Issue #8, opened 2024-02-20
Section, subsections and slide title as header level


fbob:
I know that we can use H1 and H2 header levels like this to organize a presentation : 

= Section
== Slide title

But is it possible to segment the presentation using  H1,H2 and H3 header levels like this : 

= Section
== Subsection
=== Slide title

It could make conversion from other format (I'm using obsidian+ pandoc export)  much easier.

OrangeX4:
I think it's a good idea, and I'll consider adding it in the next release, thanks :-)

Issue #7, opened 2024-02-18
Is it possible to set slide titles by level-2 headings but create slides with `#slide` function?


HPDell:
Hi! Thanks for your great package.

I've learnt that there are two modes to create packages. In one way, we use `#slide` to create slides and put the title in the argument `title`. In the other way, we show the document with `#slides` and use headings to create slides.

However, I wonder whether a third way is possible. That is to set sections, subsections, and titles with headings (of levels 1, 2, and 3 respectively), but they do not create any slides. Instead, we still use `#slide` to create slides. By this practice, we can leave a clear structure in the document. And when we need to create several slides with a same title, we don't need to type the title repeatedly.

For example

```typst
= Section 1

#new-section-slide[
  Additional content
] // -> A section slide

== Slide A

#slide[
  Some content
] // -> A slide with title "Slide A" and content "Some content".

#slide[
  Some other content
] // -> Another slide with title "Slide A" and content "Some content".
```

Contents outside the range of `#slide` are ignored.

OrangeX4:
I also really like this syntax, and I've considered it before. The most challenging of implementing this syntax might be performance issues. To achieve the desired syntax, we would need to use a syntax like `#show: slides`. However, due to triggering the execution of `slides(body)` function every time there is a small change, there could be significant performance issues, especially in cases where there is a substantial amount of content.

Additionally, we would need a way to encapsulate the `slide` function and `title-slide` with metadata.

HPDell:
How about using the section states? The headings are used to change a state, and the `slide` function choose the last section there and render the section's name as its title? Just think about some books that the current section is shown in each header (see the following image as an example). Is it possible to copy that design?

![image](https://github.com/touying-typ/touying/assets/17842185/5e81a800-daf7-4c9d-a39d-a5cd047df3ca)

The following codes are showing the idea that was implemented by myself a few months ago.

```typst
#let template(body) = {
  show heading.where(level: 3): none
  ...
  body
}

#let slide() = {
  ...
  locate(loc => query(
    selector(heading.where(level: 3)).before(loc), 
    loc
  ).last().body)
}
```

But I cannot use them with polylux, or the rendering will be broken.

OrangeX4:
I understand your idea, and I've thought about it before, but the actual problem is updating state before `#slide[]` will creates an empty page. This is due to typst's own limitations, i.e. 'set page()' creates a new page. A big part of touying's job is to avoid this problem.

```typst
[...]

PR #11, opened 2024-02-25
feat: better show-slides mode

OrangeX4:
As mentioned in #7 and #8, we plan to implement a more powerful show-slides mode, which can be broadly divided into the following three parts:

1. Using heading 1, 2, 3 to respectively correspond to section, subsection, and title. We no longer need to mix subsection and title using the second-level heading. **(This is a breaking change)**
2. Simplifying the implementation of the slides mode with themes, eliminating the need to register the `slide-in-slides` method. Instead, we adopt the "convention over configuration" philosophy to simplify the implementation. Specifically, in the slides mode, the default call is made to the `self.methods.slide` method. Additionally, if we have a `self.methods.touying-new-section-slide` method, `= Section` will be automatically translated into `self.methods.touying-new-section-slide[Section]`.
3. Through syntax like `#let (slide, focus-slide) = utils.slides(s)`, we can use syntax like `#slide[]` and `#focus-slide[]` in the show-slides mode, providing more powerful functionality and clearer structure. In this syntax, section, subsection, and title will be automatically passed into `#slide[]`. (Because `"slide"` is registered in `self.slides = ("slide",)`, it can be automatically passed in)

Here is a specific example:

```typst
#import "../lib.typ": *

#let s = themes.university.register(s)
#let s = (s.methods.info)(
  self: s,
  title: [Title],
  subtitle: [Subtitle],
  author: [Authors],
  date: datetime.today(),
  institution: [Institution],
)
#let (init, slides) = utils.methods(s)
#show: init

#let (slide, focus-slide) = utils.slides(s)
#show: slides

= Section

== Subsection

=== Title

#slide[
  AAA
]

#slide[
  BBB
]

#focus-slide[
  CCC
]
```

image

cc @fbob and @HPDell

### DIFF
--- a/slide.typ
+++ b/slide.typ
@@ -320,6 +320,7 @@
   composer: utils.side-by-side,
   section: none,
   subsection: none,
+  title: none,
   ..bodies,
 ) = {
   assert(bodies.named().len() == 0, message: "unexpected named arguments:" + repr(bodies.named().keys()))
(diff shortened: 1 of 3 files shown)


HPDell:
That is exactly what I want. Thank you!

BTW, how can I map `==` headings to slide titles? In most cases, I won't need subsections.

OrangeX4:
Do you think there should be a way to do this? is it not convenient to use `===`?

fbob:
A slide level in the header's hierarchy could be defined to specify that headings with the specified level create slides. 
See for example  [Pandoc](https://pandoc.org/chunkedhtml-demo/10.1-structuring-the-slide-show.html).
- headings above this level in the hierarchy are used to divide the presentation into sections, subsection and so on;
- headings below this level could create subheadings inside a slide. 

Regards,

HPDell:
Another quick question: shall I use the callback-style functions to use the animations like `cover` and `uncover`?

OrangeX4:
Yes, while maybe I could provide a non-callback function of `only` and `uncover` (with some limitations), I think there are two styles that make touying more difficult to understand.

OrangeX4:
I've added the slide-level parameter.  We use the `slide-level` parameter in slides to represent the structural depth. For example, the default setting for the university theme is `slide-level: 1`, where heading 1 and 2 represent section and title, respectively. If we use `#show: slides.with(slide-level: 2)`, heading 1, 2, and 3 will represent section, subsection, and title, respectively.

HPDell:
> I've added the slide-level parameter. We use the `slide-level` parameter in slides to represent the structural depth. For example, the default setting for the university theme is `slide-level: 1`, where heading 1 and 2 represent section and title, respectively. If we use `#show: slides.with(slide-level: 2)`, heading 1, 2, and 3 will represent section, subsection, and title, respectively.

So the number starts from 0 😄

OrangeX4:
Touying 0.3.1 has been released :-)

PR #19, opened 2024-03-21
docs: initialize variable s in table of contents example

pvelayudhan:
Hello,

The third example (table of contents) in the sections and subsections chapter: https://touying-typ.github.io/touying/docs/sections does not compile for me due to `error: unknown variable: s`. I think a line similar to the one I've added was missing.

The compiled PDF still doesn't exactly match the corresponding .png example. The compiled PDF in this pull request adds a gray header titled "Section" - I'm not sure if that is because simple is the wrong theme or if the .png just needs to be updated.


OrangeX4:
Thanks, these png images need to be updated.

Issue #26, opened 2024-04-03
Regression: footnotes doesn't appear since 0.3.0


sefidel:
Footnotes on the `simple` theme example doesn't render since v0.3.0.

# v0.3.0
![v0 3 0-slide](https://github.com/touying-typ/touying/assets/71049646/c623bd51-261c-4124-ae97-253303a5fa15)

# v0.2.1
![v0 2 1-slide](https://github.com/touying-typ/touying/assets/71049646/e0893409-ca9d-48e2-a8f8-8ad11cc528a2)

<details><summary>Minimal reproducible example</summary>

```typ
// #import "@preview/touying:0.2.1": *
#import "@preview/touying:0.3.0": *

#let s = themes.simple.register(s, aspect-ratio: "16-9", footer: [Simple slides])
#let s = (s.methods.enable-transparent-cover)(self: s)
#let (init, slide, slides, title-slide, centered-slide, focus-slide) = utils.methods(s)
#show: init

#title-slide[
  = Keep it simple!
  #v(2em)

  Alpha #footnote[Uni Augsburg] #h(1em)
  Bravo #footnote[Uni Bayreuth] #h(1em)
  Charlie #footnote[Uni Chemnitz] #h(1em)

  July 23
]

#slide[
  == First slide

  #lorem(20)
]

#focus-slide[
  _Focus!_

  This is very important.
]

#centered-slide(section: [Let's start a new section!])

#slide[
  == Dynamic slide
  Did you know that...

  #pause
  ...you can see the current section at the top of the slide?
]
```
</details>

OrangeX4:
Thanks, I'll take a look.

fran-miao:
> Thanks, I'll take a look.

Meet the same problem @OrangeX4

PR #27, opened 2024-04-04
refactor: remove `padding` and add `full-header`, `full-footer` to support footnote

OrangeX4:
(Original page layout ↑)

To support footnote, we need to modify the page layout of the slide from:

```typst
page(
  header: header,
  pad(
    body
  ),
  footer: footer,
)
```

to

```
set page(
  header: negative-pad(header),
  footer: negative-pad(footer),
)
body
```

structure.

Where `let negative-pad = pad.with(x: -margin.x)`, which extends the header/footer to occupy the entire width of the page.

As a result, we remove the `self.padding` parameter and add `self.full-header` and `self.full-footer` to control whether this feature is enabled (not enabled for the simple theme, but enabled for all other themes).

This PR is a breaking change, but it enables support for footnote #26 and allows normal usage of `set align(horizon)` set rule.

cc @sefidel and @fran-miao

### DIFF
--- a/slide.typ
+++ b/slide.typ
@@ -323,9 +323,43 @@
   return (result-arr, max-repetitions)
 }
 
+#let _get-negative-pad(self) = {
+  let margin = self.page-args.margin
+  if type(margin) != dictionary and type(margin) != length and type(margin) != relative {
+    return it => it
+  }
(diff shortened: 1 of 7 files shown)


OrangeX4:
image
image
image
image
image

sefidel (approved):
LGTM, tested functionality on `simple` theme.

Issue #77, opened 2024-09-13
Alert content wouldn't show in a table


HPDell:
Version info

- touying version: 0.5.2
- typst version: 0.11.1 

First of all, thank you for making this amazing package.

Today I'd like to write a table with some alert content like this

```typ
#table(
  columns: (auto,) + (1fr, ) * 6,
  align: center,
  inset: .5em,
  alert[Formula symbol], [`g1`], [`g2`], [`h1`], [`h2`], [`h3`], [`z1`],
  alert[Math symbol], [$gamma_1$], [$gamma_2$], [$beta_1$], [$beta_2$], [$beta_3$], [$mu_1$]
)
```

However, the alert contents are missing in the output table.

![image](https://github.com/user-attachments/assets/5636d130-c586-46d4-9c41-c804c0df5d78)

It should be something like this

![image](https://github.com/user-attachments/assets/a0b60bf4-afa3-42c9-8c3d-bdb7d4c53d63)

But the alert works fine outside a table.

OrangeX4:
This is a bug and I will fix it later. But maybe you might want to consider using *content* instead of #alert[content].

HPDell:
I've tired. But I'd like to show contents with the primary color. However, `_content_` would not do it for me.

PR #82, opened 2024-09-24
fix: fix no alert in table cells 

wznmickey:
Solve https://github.com/touying-typ/touying/issues/77

Previous, the items in cells will be omitted due to `continue`.

I am not quite familiar with the code. If it has conflicts with other exist functions, maybe we could add a `if` instead? 
Like
```
if kind !="touying-fn-wrapper" {
          continue
        }
```

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -910,14 +910,15 @@
         let kind = it.body.value.at("kind", default: none)
         if kind == "touying-pause" {
           repetitions += 1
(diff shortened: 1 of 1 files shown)


OrangeX4:
It is not correct to remove continue directly; instead, move continue inwards.

OrangeX4 (approved):

PR #98, opened 2024-10-15
docs: improve param documentation

Myriad-Dreamin:


OrangeX4:
How to make tinymist to support `#example()`? Syntax like

```typst
/// #example(`
/// #alternatives-match((
///   "1, 3-5": [this text has the majority],
///   "2, 6": [this is shown less often]
/// ))
/// `)
```

will make docs disappear.

OrangeX4:
Now I'm using

```typst
/// Example: `#something()`
```

Myriad-Dreamin:
We can completely follow format of tidy docs rather than workaround bugs, as we'll finally fix them. And, we can add more features if there should be missing ones.

OrangeX4 (approved):

PR #114, opened 2024-11-11
Also hide `enum` numbers with `show-hide-set-list-marker-none`

### LINKED ISSUE #10: pause with enum and list
pause behave weirdly with enum and list

image

It shouldn't show the number in early slide?

rmburg:
Fixes #10

Previously, the `show-hide-set-list-marker-none` hack only fixes lists, but enum numbers were still visible.


OrangeX4:
Thanks.

PR #120, opened 2024-11-21
fix progress bar not to break apart

enkaito:
I found an issue with the rendering of progress bars in the section slides. I have added a single line of code to prevent this problem.

Thank you for considering my contribution. Feel free to use or modify it as needed!

## minimal example
```typst
#import themes.metropolis: *
#show: metropolis-theme
#set grid(gutter: 1em)

= First chapter
= Second chapter
```

This is how the example above renders in typst.app.
You may see the progress bar is broken in the first chapter title slide.
image

### DIFF
--- a/src/components.typ
+++ b/src/components.typ
@@ -75,6 +75,7 @@
   grid(
     columns: (ratio * 100%, 1fr),
     rows: height,
+    gutter: 0pt,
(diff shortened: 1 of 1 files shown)


OrangeX4 (approved):

OrangeX4:
Thanks!

Issue #72, opened 2024-09-05
Unclear why layout did not converge within 5 attempts


pvelayudhan:
Hello,

Thank you for all the amazing recent updates you have made!

I am finding that some slides I am preparing with a combination of pauses and figures have been returning the `warning: layout did not converge within 5 attempts` message and am not exactly sure why that is. 

Here is a MWE that results in this issue:

```
#import "@preview/touying:0.5.2": *
#import themes.dewdrop: *
#show: dewdrop-theme

= A

#figure(image("square.svg", height: 40%))

#pause 1

= B

2 #pause

= C
```

I've attached square.svg as a file here:
 
![square](https://github.com/user-attachments/assets/298d9a09-d959-4e0b-9331-55c9802af9ee)

For whatever reason, any of the following adjustments renders the slides without any problems:

1. Removing the figure
2. Removing either of the lines containing pauses
3. Removing section headers B or C

Any insight as to what I may be doing wrong here or how to avoid this would be greatly appreciated!

OrangeX4:
This will be fixed in typst 0.12 and you can fix it by `config-common(enable-frozen-states-and-counters: false)` for now.

pvelayudhan:
Great to hear and that did fix it, thank you!

OrangeX4:
This issue should have been fixed in Typst 0.12 and I'll close this issue.

nycex:
I do still have an (at least similar) issue with typst `0.12.0`:

```typst
#import "@preview/touying:0.5.3": *
#import themes.metropolis: *

#show: metropolis-theme.with(
  // config-common(enable-frozen-states-and-counters: false)
)

#figure(
  "B",
)
#pause

- U

#pagebreak()

+ V
#pause

W

#pagebreak()

Y
#pause
Z

```


Here, I also get a `layout did not converge within 5 attempts` that is _fixed_ by `config-common(enable-frozen-states-and-counters: false)`.

aparatext:
I had the same issue in my presentation. typst 0.14.2, touying 0.7.0.

`enable-frozen-states-and-counters: false` fixed that.

hashcatHitman:
Seems like the same issue for me too. Typst 0.14.2, touying 0.7.3. Minimal example included.

```typst
#import "@preview/touying:0.7.3": *
#import themes.simple: *
#show: simple-theme

#slide[
  #pause
][
  #figure(
    none, // This could have content, it doesn't seem to matter
  )
]
#pause
[...]

PR #124, opened 2024-11-28
Fix frozen-counters bug with multiple #pause commands

enkaito:
## Description

This PR addresses a bug in the frozen-counters functionality where status updates show unexpected behavior when using multiple `#pause` commands. This issue may be related to #72.

## Problem

When using multiple `#pause` commands, figure captions are not correctly incremented. For example, in the example below with multiple figures and pauses, the final figure might be captioned as "Figure 2:" though it should be "Figure 3:".

```typst
#import themes.simple: *

#show: simple-theme

== slide 1
#figure([test figure], caption: [caption])

== slide 2
#figure([test figure], caption: [caption])
#pause
paused

== slide 3
#figure([test figure], caption: [caption])
#pause
paused

```

## Solution

Instead of storing the counter value at the beginning of a sequence of subslides, we now store the location. We then use `status.at(location)` to rewind the status when needed. This approach ensures that counters are correctly updated across pauses and slides.

## Testing

I have tested my code with examples given in the examples folder in this repository, besides the example above, and two examples in #72. All of them compiled with no errors / warnings.

---

- As I'm not fluent in Typst, my code might not be fully ideomatic. I welcome any revisions or improvements to the implementation.
- It might be better to store the location in some outer variable (e.g. `#utils.current-slide`) and update it when creating a new slide, instead of querying the label for every new subslide.
- minor fix: I found a variable named `current`, which is not used, so I deleted it.

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -1530,6 +1530,11 @@
   (header, footer)
 }
 
+#let _rewind-states(states, location) = {
+  for s in states {
+    s.update(s.at(selector(location)))
+  }
+}
 
 /// Touying slide function, the core function of touying. It usually is used to create a slide with animation effects and works with `touying-slide-wrapper` function.
 ///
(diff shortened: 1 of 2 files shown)


enkaito:
> It might be better to store the location in some outer variable (e.g. #utils.current-slide) and update it when creating a new slide, instead of querying the label for every new subslide.

I found this idea won't work because Typst doesn't allow changing outer variables within a context. Therefore, querying the label for every new subslide is necessary with this approach.

enkaito:
I added a new state #utils.loc-prior-newslide to store the location of parent slides. This simplifies the code and improves readability (and performance?) The implementation works well in testing. Sorry for sending multiple comments in a short period.

OrangeX4:
Thank! I'll check it out later.

OrangeX4 (approved):

OrangeX4:
Thanks, I think your method is more elegant.

OrangeX4:
With this PR, we can also add the `counter(heading)` to the frozon-counters: https://github.com/touying-typ/touying/commit/34bfe9e5165923baca950339e2bd19631cf4c806

Issue #121, opened 2024-11-24
Incorrect page numbering when handout mode is on and if the page has `#pause`


anormalperson8:
Given the following code:
```typ
#import "@preview/touying:0.5.3": *
#import themes.university: *

#show: university-theme.with(
  aspect-ratio: "16-9",
  progress-bar: false,
  navigation: none,
  header-right: none,
  config-info(
    title: [Title],
    short-title: [short title],
    subtitle: [subtitle],
    author: [author],
    date: [date],
    institution: [Some place]
  ),
  config-common(
    show-hide-set-list-marker-none: true,
    handout: true // THIS LINE
  ),
)

#title-slide()
= Section 1
== Page 2
Testing
#pause
Testing 2
#pause
Testing 3

== Page 3
Testing
#pause
Testing 2

== Page 4
Testing
#pause
Testing 2

= Section 2
== Page 6
Testing
#pause
Testing 2

= Section 3
== Page 8
Testing

== Page 9
Testing
```
If `handout` is `true` and the slide after a section (single `=`) uses `#pause`, then the page numbering on those pages would be completely broken.

The title of each subsection should refer to its actual page number, but compiling this with Typst 0.12.0 makes them wonky.

I hope I didn't configure something incorrectly. Thanks!

PR #125, opened 2024-11-28
fix: incorrect page num when draft is true

enkaito:
I removed one line to correctly handle page numbering when draft mode is true. This change resolves issue #121.
I've tested it to ensure pagination works as expected in draft mode with the example given in #121.
Ready for review!


OrangeX4:
I didn't think it should break the consistency between normal mode and handout mode, so I made a small change.

OrangeX4 (approved):

PR #126, opened 2024-12-01
fix: metropolis slide's header to return content if title is specified

enkaito:
I discovered a bug in the metropolis theme where the header was returning the function itself instead of calling it. This causes an error when trying to specify the slide title.

## minimal working example

```typst
#import themes.metropolis: *

#show: metropolis-theme

#slide(title: "section with a title")[
  error!
]
```

### DIFF
--- a/themes/metropolis.typ
+++ b/themes/metropolis.typ
@@ -52,7 +52,7 @@
     components.left-and-right(
       {
         if title != auto {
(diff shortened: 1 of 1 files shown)


OrangeX4:
```typst
#import themes.metropolis: *

#show: metropolis-theme

== section with a title

content
```

OrangeX4 (approved):

Issue #112, opened 2024-11-04
Semi-transparent-cover doesn't cover everything and renders wrongly


JoshuaLampert:
You can find an MWE at : https://typst.app/project/rO30herYanvASbIkJdjjW2.
This issue has two (maybe related?) sub-issues:
1. In the first subslide we see that the lower line of the equation is not central, the "x" is not in math font, and the vector is smaller than in should be.
2. Not everything is covered. Only a strip, which is not high enough covers the central part of the equation, but the upper and lower parts are still visible without transparency.

OrangeX4:
This is due to the fact that currently semi-transparent cover is very hacky and may be difficult to fix. Maybe we can consider changing the font color instead of using a semi-rect

zral0kh:
well it should be fixed soon with #328 now being merged.

that was quite the effort and its still not perfect due to typst internal stuff, but it should suffice for most use cases.
The `color-changing-cover` and `alpha-changing-cover` can now also revert to that as fallbacks for covering blocks properly.

Note: I did somehow think we could not extract the alignment and reconstruct those things like figure, place, etc. but ofc we can. 
If you would want to properly do that we should also go into those functions and only cover the individual elements inside.
But then to use semi-rect as fallback for the text-color changing covers we also need to track that...
I will put that onto my todo list. but will likely take a while.

JoshuaLampert:
What is the status of this issue? I updated the MWE from above to use the latest touying version and as you can see, now the equations are not covered at all. Is that expected or do I need to change anything, @zral0kh?

zral0kh:
Yes that is expected as of rn. i will have to do some more stuff to properly cover equations since sth like vectors don't expand the line height etc. and even figures are not covered were nicely right now as the entire block is covered with one rect, not the image once and the text extra.

i am planning to fix this this week bc i cannot do much in the coming month

zral0kh:
so i fiddled a bit around and i have come to the conclusion that either we need to restructure quite a lot and pass sth through that we don't really want, or we deprecate cover with rect and only really use it for graphics elements and favuor the `alpha-changing-cover`. we cannot reliably do equations or list markers, or rather it's not easily achievable. with alpha-changing-cover it is, since the markers are text and equations are rendered internally via text.

[...]

PR #129, opened 2024-12-03
feat: add cover color-changing-cover

enkaito:
This pull request introduces a new (pseudo) semitransparent cover function that provides a method for hiding content in slides. The function changes the text color of elements to a specified color if they contain text. If an element does not contain text, it can use `hide` as a fallback.

This implementation has been successfully tested with the default examples and the specific example provided in issue #112.

### DIFF
--- a/src/utils.typ
+++ b/src/utils.typ
@@ -816,6 +816,9 @@
   if it.has("base") {
     return _contains-text(it.base)
   }
+  if it.has("child") {
+    return _contains-text(it.child)
+  }
   if it.has("children") {
     for child in it.children {
       if _contains-text(child) {
--- a/src/utils.typ
+++ b/src/utils.typ
@@ -803,7 +803,7 @@
 }
 
 // recursively checks if `it` has a text in it
-#let _contains-text(it) = {
+#let _contains-text(it, transparentize-table) = {
   if type(it) != content {
     return false
   }
@@ -811,17 +811,20 @@
     return true
   }
   if it.has("body") {
-    return _contains-text(it.body)
+    return _contains-text(it.body, transparentize-table)
   }
   if it.has("base") {
-    return _contains-text(it.base)
+    return _contains-text(it.base, transparentize-table)
   }
   if it.has("child") {
-    return _contains-text(it.child)
+    return _contains-text(it.child, transparentize-table)
   }
   if it.has("children") {
+    if it.func() == table and transparentize-table {
+      return true
+    }
     for child in it.children {
-      if _contains-text(child) {
+      if _contains-text(child, transparentize-table) {
         return true
       }
     }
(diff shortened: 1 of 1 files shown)


OrangeX4:
Can we achieve translucency by modifying the alpha value instead of fixing a color? We can think about the hacky code that I have as proof of concept below

```typst
#set page(width: 20em, height: auto)

#let update-alpha(constructor: rgb, color, alpha) = constructor(..color.components(alpha: true).slice(0, -1), alpha)

#show par: it => {
  if repr(text.fill).split("(").at(0) == "rgb" {
    set text(fill: update-alpha(text.fill, 50%))
    it
  } else {
    it
  }
}

// ----------------
// content
// ----------------


#set text(rgb("#000000"))

Test

#set text(rgb("#ff0000"))

Test
```

image

enkaito:
How about this?

src/utils.typ
```diff
@@ -802,6 +802,56 @@
...
+  if it.has("body") {
+    return _contains-text(it.body)
+  }
+  if it.has("base") {
+    return _contains-text(it.base)
+  }
+  if it.has("child") {
+    return _contains-text(it.child)
+  }
+  if it.has("children") {
```
OrangeX4, line 822: The `table` and `grid` types also have `children`, I'm not sure if this should be included?

enklht: I understand. It might be good to have an option to toggle the behaviour, or exclude those elements (using `util.is-sequence`?)

enkaito:
**Quick Update:** I discovered that the new `alpha-changing-cover` doesn't function perfectly when the text color is modified within the paused items.

src/utils.typ
OrangeX4: Since the table has borders, we should use `util.is-sequence` as default, but it can indeed be made into a configuration.

enkaito:
**Update:** The enhanced function now provides improved handling of color-changing texts. This is achieved by utilizing the `context` feature and the `update-alpha` function. This works well with the example I made the other day.

## Note
The decision to use `update-alpha` instead of `transparentize` was deliberate. Unlike `transparentize`, which applies relative changes that can accumulate to cause nested texts to vanish, `update-alpha` allows to specify absolute text opacities. This approach ensures consistent visibility for nested color-changing elements.

enkaito:
While this is slightly tangential to the current PR, I'd like to propose an improvement to the `update-alpha` function:

```typst
#let update-alpha(color, alpha) = color.opacify(100%).transparentize(alpha)
```

This revised implementation can handle all color constructors uniformly. It first ensures full opacity with `opacify(100%)`, then applies the desired transparency using `transparentize(alpha)`.

OrangeX4:
If `update-alpha` is used, we need to consider non-rgb cases, like `#set text(luma(80%))`.

enkaito:
My latest proposal addresses that issue!

OrangeX4 (approved):

OrangeX4:
Thanks!

Issue #109, opened 2024-11-01
`utils.fit-to-height` no longer working on typst 0.12


PhotonQuantum:
`utils.fit-to-height` produces a blank area with no content on typst 0.12. The layout is calculated correctly, though.

I bisected Typst, and it seems that it's this PR https://github.com/typst/typst/pull/4212 that introduced the bug.

In the PR description it says,

> The exact semantics of how tags will stick to elements are also still somewhat in-flux and will likely a change a bit in later refactor phases. These rules affect the results of `.page()` and `.position()` calls on `location`s. With this PR, things should mostly be the same as on main, though there will be edge cases where things changed a bit. They might change again with further refactoring.

There are other libraries affected by this change (https://github.com/typst/typst/issues/4302).

## POC

```typ
#import "@preview/touying:0.5.3": *
#import themes.simple: *

#show: simple-theme.with(aspect-ratio: "16-9")

= Title

== First Slide

#utils.fit-to-height(0.5fr)[
  #lorem(50)
]

#lorem(50)
```

## Expected

<img width="600" alt="Screenshot 2024-11-01 at 4 06 07 PM" src="https://github.com/user-attachments/assets/4f5fd474-f3fa-43ff-abc4-0752514e7f08">

## Got

<img width="600" alt="Screenshot 2024-11-01 at 4 08 29 PM" src="https://github.com/user-attachments/assets/318f33dd-8288-4c32-98a9-d1839677b6a8">

jiucenglou:
It is often useful to fit content within a single slide by automatic scaling.
I am wondering if there is any workaround at the moment ?

PhotonQuantum:
> It is often useful to fit content within a single slide by automatic scaling. I am wondering if there is any workaround at the moment ?

I haven't made a package yet, but you can check out [this code](https://lightquantum.me/lr_prose.zip) for reference. I It's a temporary solution, but I think it works well for most cases.

You can experiment with the `debug`, `precise-size` and `guess` theme options, edit the content, and see how it works.

jiucenglou:
> > It is often useful to fit content within a single slide by automatic scaling. I am wondering if there is any workaround at the moment ?
> 
> I haven't made a package yet, but you can check out [this code](https://lightquantum.me/lr_prose.zip) for reference. It's a temporary solution, but I think it works well for most cases.
> 
> You can experiment with the `debug`, `precise-size` and `guess` theme options, edit the content, and see how it works.

Thank You very much ! I will try your solution and report back.

PR #131, opened 2024-12-04
fix: behaviors of `fit-to-height` and `fit-to-width`

enkaito:
This PR addresses the behaviors of the `fit-to-height` and `fit-to-width` functions, resolving issues related to #109.

### Changes

1. **Changed to use the `width` and `height` parameters of `measure`:** Replaced previous hacks using hidden labels with `measure`'s width and height parameters to accurately determine relative sizes within a container.
2. **Removed nested box for `scale`:** Eliminated the need for a nested box to handle the size of the `scale` by utilizing the `reflow` parameter of `scale`.
3. **Updated examples in the comment:** The actual size of `1fr` is determined by other `fr`s in the same container, so only one `0.5fr` in the container is equivalent to `100%`. To avoid this confusion, I slightly changed the documented examples to use `%` instead.

### Notes

- In simple theme, the headings are rendered in slide preamble instead of in the header. This causes relative sizes in simple-theme to appear larger than expected.

---

Please review the changes, particularly the updated example and the new implementation using `measure`. Any feedback is welcome!

### DIFF
--- a/src/utils.typ
+++ b/src/utils.typ
@@ -600,42 +600,9 @@
   height,
   body,
 ) = {
-  // Place two labels with the requested vertical separation to be able to
-  // measure their vertical distance in pt.
-  // Using this approach instead of using `measure` allows us to accept fractions
-  // like `1fr` as well.
-  // The label must be attached to content, so we use a show rule that doesn't
-  // display anything as the anchor.
-  let before-label = label("touying-fit-height-before")
-  let after-label = label("touying-fit-height-after")
-  [
-    #show before-label: none
-    #show after-label: none
-    #v(1em)
-    hidden#before-label
-    #v(height)
-    hidden#after-label
-  ]
-
   context {
-    let before = query(selector(before-label).before(here()))
-    let before-pos = before.last().location().position()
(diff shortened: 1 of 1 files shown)


OrangeX4:
It seems that the current method does not solve the problem #109. What I am thinking is whether we can make the following code work properly:

```typst
#import themes.simple: *

= Title

== First Slide

#utils.fit-to-remaining-height[#lorem(50)]

#lorem(50)
```

It seems that this PR does not solve this problem at present, even using `#utils.fit-to-height(100%)[#lorem(50)]`.

Maybe we can find inspiration here: https://forum.typst.app/t/how-to-auto-size-text-and-images/1290

OrangeX4:
Thanks. Yes it solves part of the problem, I just thought maybe there is a way to solve all of it. Maybe we still need to think about it.

enkaito:
By the way, have you tried the new `fit-to-height` function with a small percentage parameter, like 25%? I think this is the intended behavior in the example of issue #109. Since the parameter-free function is currently too complex, it makes sense to close the bug-fix issue and open a new one for `fit-to-remaining-height`. What do you think?

PR #134, opened 2024-12-08
fix: do not hardcode page sizes

t1mlange:
Hello,

I had a use-case for a different page size (i.e. higher dpi). However, manually providing width and height to the page broke the speaker notes feature. Just replacing the hard-coded values with the actual page size seems to just work fine.


OrangeX4 (approved):

OrangeX4:
Thanks.

Issue #111, opened 2024-11-04
Color text on set of subslides


JoshuaLampert:
First of all, thank you for the very nice package. I recently built my own template based on my previous LaTeX template and managed to create almost identical results (while benefiting from typst and touying compared to LaTeX and beamer).
I mainly have two issues, which currently hinder me in completely switching to typst for presentations (this issue and an upcoming one).
I usually heavily use coloring of certain parts (in LaTeX with `\color<>[]{}`), i.e. I want some text (or part of a formula) to be colored _only in a set of subslides_. If I haven't missed anything, this is currently not supported by touying out-of-box. After looking at the source code, I figured out I can hack something together by defining it on my own. See https://typst.app/project/r2RWPmJouYepvYdkL3EtPE for an MWE. This already works quite nicely in simple situations, but doesn't work for more complex ones, where I need the callback style because IIUC I cannot get my `colors` function by `utils.methods(self)`. Note that my example works without the `semi-transparent-cover`, which I don't quite understand.
Is there interest in including such a function in touying? If yes, I would be happy to create a PR with an initial suggestion.

OrangeX4:
This is because `semi-transparent-cover` will place paused contents into a context expression, making `touying-fn-wrapper` cannot working properly. However, you can use `config-methods(color: my color)` to add it to methods and obtains it by `#let (uncover, color) = utils.methods(self)`.

```typst
#import "@preview/touying:0.5.3": *
#import themes.simple: *

#let my_color(self: none, visible-subslides, col, cont) = {
  if utils.check-visible(self.subslide, visible-subslides) {
    text(fill: col, cont)
  } else {
    cont
  }
}

#let color(visible-subslides, col, cont) = {
  touying-fn-wrapper(
    my_color,
    last-subslide: utils.last-required-subslide(visible-subslides),
    visible-subslides,
    col,
    cont,
  )
}

#show: simple-theme.with(
  aspect-ratio: "16-9",
  config-methods(
    cover: utils.semi-transparent-cover.with(alpha: 85%),
    color: my_color,
  ),
)

= Title

== Title of Slide

#color("2-", red, "this works as I want")

#slide(repeat: 3, self => [
  #let (uncover, color) = utils.methods(self)
  This doesn't:

  $ f(x) &= (x + 1)^2 \
    uncover("2-", color("2", #red, &= x^2 + x + x + 1)) \
    uncover("3", &= x^2 + 2x + 1)
  $
])
```

[...]

PR #139, opened 2024-12-18
Reuse effect from utils

JoshuaLampert:
See https://github.com/touying-typ/touying/issues/111. Could we reuse the definition of `effect` from utils.typ in core.typ? Right now, `effect` from utils.typ doesn't get an argument `is-method`.

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -516,7 +516,7 @@
     utils.effect,
     last-subslide: utils.last-required-subslide(visible-subslides),
     visible-subslides,
-    is-method,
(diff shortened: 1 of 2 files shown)


JoshuaLampert:
Ok, I tested this now and it works for me as I would expect. Can you maybe have a look @OrangeX4? Maybe we can still include this in 0.5.5 by committing this patch [here](https://github.com/typst/packages/pull/1446)?

OrangeX4 (approved):

PR #145, opened 2025-01-16
add syrkis theme

syrkis:
Adding scandinavian styled theme with two column intro page (title and toc).

[slide.pdf](https://github.com/user-attachments/files/18431726/slide.pdf)


OrangeX4:
Thank you for your contribution. I think this theme is pretty good, but if it is directly used as a built-in theme, it will increase the complexity of maintenance, so I don't plan to merge this theme for the time being. You can consider uploading it to typst universe, thank you!

PR #153, opened 2025-02-17
Fix type string comparison

eltos:
Context: https://github.com/typst/packages/issues/1725


OrangeX4 (approved):

OrangeX4:
Thanks, I thought I had fixed it all...

OrangeX4:
The good news is that 99% of the time touying doesn't go through this code, so it doesn't have much of an impact.

PR #157, opened 2025-02-27
fix: show-hide-set-list-marker-none with full enum

BuriedInTheGround:
Animation effects such as `#pause` make `set enum(numbering: n => none)` throw an `unexpected argument` error when `#set enum(full: true)` is applied. This PR fixes that.


OrangeX4 (approved):

OrangeX4:
Thanks!

PR #161, opened 2025-03-05
Allow customisation of components.checkerboard

elinscott:
Adding options to change the currently hard-coded background color and text alignment within `component.checkerboard`

### DIFF
--- a/src/components.typ
+++ b/src/components.typ
@@ -112,11 +112,11 @@
 ///
 /// That means that `#checkerboard[...][...]` stacks horizontally and `#checkerboard(columns: 1)[...][...]` stacks vertically.
 /// 
-/// - alignment (alignment): The alignment of the contents of each checkerboard cell (default: center + horizon)
+/// - alignment (alignment): The alignment applied to the contents of each checkerboard cell. Default is `center + horizon`.
 /// 
-/// - primary (color): The background color of odd cells (default: white)
+/// - primary (color): The background color of odd cells. Default is `white`.
 /// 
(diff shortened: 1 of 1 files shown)


src/components.typ
```diff
@@ -113,7 +113,7 @@
 /// That means that `#checkerboard[...][...]` stacks horizontally and `#checkerboard(columns: 1)[...][...]` stacks vertically.
 ///
 /// -> content
-#let checkerboard(columns: none, rows: none, ..bodies) = {
+#let checkerboard(columns: none, rows: none, alignment: center + horizon, color1: white, color2: silver, ..bodies) = {
```
OrangeX4: Maybe we can use primary-color and secondary-color as names

src/components.typ
```diff
@@ -113,7 +113,7 @@
 /// That means that `#checkerboard[...][...]` stacks horizontally and `#checkerboard(columns: 1)[...][...]` stacks vertically.
```
OrangeX4, line 113: The new parameters should be explained in the comments

src/components.typ
elinscott: I changed it to simply `primary` and `secondary` following the example of `progress-bar`, `simple-navigation` and `knob-marker`. I hope this is OK!

src/components.typ
elinscott: Done :slightly_smiling_face:

OrangeX4 (approved):

OrangeX4:
Thanks!

PR #167, opened 2025-03-24
fix: use regex to override colors of equations

### LINKED ISSUE #156: color-changing-cover not working in equations
Hi,

thanks for the work put into this package!

While preparing a presentation I noticed the following problem: the code introduced in #129 works very good in general, even when the color of the text is changed. However, when the color of a variable is changed inside an equation, this does not work, i.e. the color is visible with 100% opacity when it should be grey/transparent.

So, if I have this:

```typst
#slide[
    - something
    #pause
    - equation with color
      $
       text(fill: red, F) = m dot a
      $
]
```

then the `F` will be colored red the whole time. Is there a solution for this?

enklht:
Currently, `color-changing-cover` and `alpha-changing-cover` cannot override the colors of equations correctly.
I found that we can use the regex `.+` to match every string and override the colors of equations.

Thanks.

Closes #156

### DIFF
--- a/src/utils.typ
+++ b/src/utils.typ
@@ -361,7 +361,7 @@
 /// - setting (function): The setting of the heading. Default is `body => body`.
 ///
 /// - style (function): The style of the heading. If `style` is a function, it will use the function to style the heading. For example, `style: current-heading => current-heading.body`.
-/// 
+///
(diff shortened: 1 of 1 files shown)


OrangeX4 (approved):

OrangeX4:
Thanks.

PR #171, opened 2025-04-04
fix: use correct circle symbol

jthulhu:
`sym.circle` creates a big, hollow circle whereas `sym.circle.fill` creates a small, filled circle.  Replaced `sym.circle` with `sym.circle.small` to have consistent circle size.


OrangeX4 (approved):

OrangeX4:
Thanks.

PR #207, opened 2025-08-07
feat: Pass named arguments of side-by-side to grid.

cooki35:
The update to the side-by-side component passes the named arguments further down to the grid. This allows us to pass e.g., ```align:top``` to side-by-side and in turn to grid to align the columns.

Another example would be stroke:
```typst
#side-by-side(stroke: 1pt+red)[
Test123
][
Test1
Test2
Test3
]
````
Currently this would result in side-by-side ignoring the stroke argument.
With the fix the columns would have red borders.

### DIFF
--- a/src/components.typ
+++ b/src/components.typ
@@ -22,6 +22,7 @@
 ///
 /// -> content
 #let side-by-side(columns: auto, gutter: 1em, ..bodies) = {
+  let args = bodies.named()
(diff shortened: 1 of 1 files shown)


OrangeX4 (approved):

OrangeX4:
Thanks!

PR #218, opened 2025-09-17
Parameterize authors per row in university theme

yagehu:
For example, if you have eight authors, it's prettier to set authors per row to 4 instead of the default 3.

### DIFF
--- a/themes/university.typ
+++ b/themes/university.typ
@@ -129,6 +129,13 @@
       (authors,)
     }
   }
+
+  let authors-per-row = if "authors-per-row" in info {
+    info.authors-per-row
+  } else {
+    3
+  }
+
   let body = {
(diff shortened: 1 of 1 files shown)


themes/university.typ
```diff
@@ -129,6 +129,13 @@
       (authors,)
     }
   }
+
+  let authors-per-row = if "authors-per-row" in info {
```
OrangeX4, line 133: Maybe consider passing in as an arg instead of info?

yagehu: It can currently be used as:

```typst
#title-slide(authors-per-row: 4)
```

`info` comes from `args`:

```typst
let info = self.info + args.named()
```

What do you mean by passing in as an arg? Sorry I'm new to Typst.

PR #228, opened 2025-10-26
feat: minislides can be displayed inline

5p4k:
I was porting some Beamer theme to typst, and in the source theme I was using minislides, but the bullets were displayed right after the section title to save vertical space. I patched `mini-slides` locally, but I thought this feature might be interesting for other users too.

Here's the look I was after:

20251026_19h51m36s_grim

And here is how the dewdrop theme would look with `inline: true, linebreaks: false`:

20251026_19h51m57s_grim

I'm not very experienced with typst, so I'm looking for feedback on the change -- for example, I'm not convinced that two settings with similar name (`linebreaks` and `inline`) should exist for this, perhaps it should be better to have `linebreaks` be some sort of enum, like "always", "inline-subsections", "inline-all".

Let me know what you think and if this feature is interesting for touying.

### DIFF
--- a/src/components.typ
+++ b/src/components.typ
@@ -424,6 +424,8 @@
 ///
 /// - short-heading (boolean): Indicates whether the headings should be shortened. Default is `true`.
 ///
+/// - inline (boolean): Indicates whether the bullets are displayed right after the text, instead of breaking the line. Default is `false`.
+///
 /// -> content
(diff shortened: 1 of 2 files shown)


OrangeX4:
I think it's good, sorry I only have time to see it now, thanks!

PR #237, opened 2025-11-27
fix: theorion package is broken with Typst 0.14.0

Porkepix:
Version 0.4.0 is broken and doesn't compile with Typst 0.14.0. This was fixed in the next version 0.4.1 according to https://github.com/OrangeX4/typst-theorion/releases/tag/0.4.1 and https://github.com/OrangeX4/typst-theorion/commit/b3364f8b33d6d06290bce828bea06133fbeb75c9

Example of the breakage with example from repo:

```
error: unexpected argument
    ┌─ @preview/theorion:0.4.0/core.typ:298:20
    │
298 │   let display-number(get-loc: here) = (counter: frame-counter, ..args) => context {
    │                     ^^^^^^^^^^^^^^^
```


OrangeX4:
Thanks

Issue #229, opened 2025-10-27
Bibliography compile error in Typst 0.14


frabera:
In the new version of Typst (0.14) adding a bibliography produces a compilation error:

MWE
```typst
#import "@preview/touying:0.6.1": *
#import themes.simple: *

#let biblio = bytes(
  "@book{dirac,
    title={The Principles of Quantum Mechanics},
    author={Paul Adrien Maurice Dirac},
    series={International series of monographs on physics},
    year={1981},
    publisher={Clarendon Press},
  }",
)

#show: simple-theme.with(
  aspect-ratio: "16-9",
  config-common(show-bibliography-as-footnote: bibliography(biblio)),
)

= Title

== First Slide

Hello, Touying! @dirac

---

#magic.bibliography()
```

Error:
```
error: array index out of bounds (index: 0, len: 0) and no default value was specified
    ┌─ @preview/touying:0.6.1\src\magic.typ:173:26
    │
173 │             let bibitem = bibliography-state.final().at(bibliography-counter.get().at(0))
    │                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

In Typst 0.13.1, this is the result:

<img width="811" height="915" alt="Image" src="https://github.com/user-attachments/assets/13009157-f585-4b65-9630-3dede7767192" />

This is the method I was using to generate citations as footnotes. In alternative, the following method does not produce compilation errors, but the citations are not rendered:

MWE
```typst
#import "@preview/touying:0.6.1": *
#import themes.simple: *

#let biblio = bytes(
  "@book{dirac,
    title={The Principles of Quantum Mechanics},
    author={Paul Adrien Maurice Dirac},
    series={International series of monographs on physics},
    year={1981},
    publisher={Clarendon Press},
  }",
)

#show: simple-theme.with(
  aspect-ratio: "16-9",
)

#show: magic.bibliography-as-footnote.with(bibliography(biblio))

= Title

== First Slide

// #show: bibliography-as-footnote.with(bibliography(biblio))

Hello, Touying! @dirac

== Bibliography

#magic.bibliography(title: none)
```

Result (in both Typst 0.13.1 and 0.14):

<img width="816" height="924" alt="Image" src="https://github.com/user-attachments/assets/560de431-5212-49f1-8611-c955d0467bab" />

frabera:
[...]

PR #249, opened 2025-12-13
fix magic.bibliography-as-footnote in typst 0.14

AngelPone:
This pull request fixes `magic.bibliography-as-footnote` (#229) through the following fixes:  
1. extract the bibliography using `cite(<key>, form: full)`
2. use `regex` to remove `[1]` at the beginning of rendered citation. 
3. record the bibitem everytime a new key is visited.
4. reuse the footnote if some item is cited multiple times in one subsection.

Based on this new mecanism, the state `bibliography-map` and `bibliography-counter`, and the argument `record` will be unnecessary and thus are removed.


jhalmen:
I'd tried this quickly last week, and the footnote citations only appear on the first sub-slide when using animations i.e. `#uncover` and family

AngelPone:
@jhalmen I've pushed some changes to address this issue, and it should now be resolved in most scenarios.

However, during testing, I noticed a couple of edge cases that seem to be outside the scope of this PR:
1. **Footnote visibility with animations**: Citations appear in the footnotes immediately, even if the #cite has not yet been revealed. For example:
```typ
- Item 1 @some-key  #pause
- Item 2 @another-key
```
On the first slide, only the bibliography entry for Item 1 should be visible. However, entries for both @some-key and @another-key are displayed. This seems to be related to how Touying handles animation internally.

2. **Page overflow with animations**: When content is long enough to overflow to a second page, combining this with animation can sometimes result in blank pages.

Note: Both issues were also present in the old bibliography-as-footnote implementation (pre-Typst 0.14.0), so they are likely pre-existing issues rather than regressions introduced by this PR.

martenmatrix:
Tysm, @AngelPone

OrangeX4 (approved):

OrangeX4:
Sorry I was too busy to review this PR. It has been merged now, and I will release it as soon as possible. Thank you!

PR #256, opened 2026-01-23
fix: support ratio and relative margins for full-width headers

### LINKED ISSUE #203: Layout error in themes other than simple if a value of type `ratio` are passed to the `margin` argument of `config-page`
Hi,

Firstly, great thanks to your wonderful package.

I'm making slides in Chinese, and to avoid things like punctuation getting squished or the text looking all over the place, I try to keep the content area a nice even width—like a whole number of characters wide. Since Typst has some limitations right now, I do this by setting the left and right margins to `(100% - 1em * n) / 2`, where *n* is how many characters I want to fit. But when I use certain themes, the layout starts to break.

For example,

```typ
#show: university-theme.with(
  config-info(
    title: "测试 Touying 幻灯片",
    date: datetime.today(),
    institution: [单位],
    logo: emoji.school,
  ),
  config-page(
    margin: (top: 2.25em, bottom: 1em, x: (100% - 30em) / 2),
  )
)
```

Results in

Image

But setting the x margin to relative lengths (like 2em) can result in correct layout.

Image

I think this is a bug, because in the core code of Touying, the margin handling only covers values of type `length` and `relative`, but doesn't account for those of type `ratio`.

Lyrics2196:
Previously, `_get-negative-pad` relied on `pad` to compensate for page margins using the same margin value. However, `pad` resolves percentages relative to the content width, while page margins are relative to the page width. This caused incorrect layout (too narrow headers/footers) when `config-page` used `ratio` or `relative` types for margins. 

This commit updates `src/core.typ` to explicitly convert `ratio` and `relative` margin values to absolute lengths based on the page width before applying negative padding.

Fixes #203


src/core.typ
```diff
@@ -1884,33 +1884,56 @@
...
-  if "left" in margin {
-    pad-args.left = -margin.left
-  }
-  if "right" in margin {
-    pad-args.right = -margin.right
-  }
-  if "rest" in margin {
-    pad-args.x = -margin.rest
+
+  return it => context {
```
OrangeX4, line 1900: Is the context necessary here? If so, can its impact be minimized? That is, should this processing only be applied to ratios and relative margins?

Lyrics2196: Yes, the `context` is necessary. Because the `page.width` can only be uesed when context is known. I took care to minimize modifications. Since margin could be `dictionary`, applying changes only when margin or its element type is `ratio` or `relative` would require additional code.

OrangeX4: If the margin is dictionary type, we should first check if this dictionary contains a value of type ratio. If not, we should be able to use the old logic.

src/core.typ
OrangeX4, line 1: Need typstyle for formatting.

src/core.typ
Lyrics2196: OK. I added a check for when the margin is a dictionary, calling `to-abs()` only when the elements within the margin are relative or ratio. I attempted to reduce the scope of the context wrapper, but it failed. This is because all content that calls `page.width` must be wrapped in the context, and the parameters of the pad function cannot be context, which prevents me from wrapping only `to-abs()` within the context.

OrangeX4:
Thanks!

Issue #263, opened 2026-03-03
[Stargazer theme] What does `<touying:skip>` do in theme `stargazer`?


Fenyutanchan:
### MWE:
```typst
#import "@preview/touying:0.6.1": *
  #import themes.stargazer: *
#import "@preview/numbly:0.1.0": *

#show: stargazer-theme.with(
  aspect-ratio: "4-3",
  config-info(
    title: [TT],
    subtitle: [tt],
    author: [xxx],
    date: [],
    institution: [yyy],
  ),
)

#set heading(numbering: numbly("{1}.", default: "1.1"))

#title-slide()

#outline-slide()

= Something <touying:skip>

?

```

### Output:
<img width="971" height="1415" alt="Image" src="https://github.com/user-attachments/assets/64267448-bf5f-46a4-83e0-5a0659efe7e0" />

### Expectation:
"1. Something" instead of "1. Skip" in the outline!

PR #265, opened 2026-03-04
fix: skip special touying labels in short-heading function

Fenyutanchan:
Handle special touying labels (touying:hidden, touying:skip, touying:unnumbered, touying:unoutlined, touying:unbookmarked) in the short-heading function by returning the raw body text instead of converting them to short heading format.

Fix for #263

### DIFF
--- a/src/utils.typ
+++ b/src/utils.typ
@@ -346,6 +346,15 @@
     return it.body
   }
   let lbl = str(it.label)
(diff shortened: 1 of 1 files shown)


Fenyutanchan:
Hi, @OrangeX4 

Could you review this fix?

OrangeX4:
Thanks

PR #247, opened 2025-12-11
Expose `step`, `at`, `until` and `between`

wvhulle:
For usage by Fletcher to hide elements after show:

```typst
#import "/lib.typ": *
#import themes.university: *
#import "@preview/fletcher:0.5.8" as fletcher: edge, node

#let fletcher-diagram = touying-reducer.with(
  reduce: fletcher.diagram,
  cover: fletcher.hide,
)

#show: university-theme.with(aspect-ratio: "16-9")

== Show in range with `between`

#slide[
  Content visible only during specific slide ranges:

  #fletcher-diagram(
    node-stroke: .1em,
    spacing: 4em,
    // Always visible
    node((0, 0), [Start], radius: 2em),
    node((3, 0), [End], radius: 2em),
    // Blue edge visible only on slides 2-3
    between(2, 3, edge((0, 0), (1, 0), "-|>", stroke: blue, label: [2-3])),
    // Middle node visible on slides 2-4
    between(2, 4, node((1, 0), [Mid], radius: 2em, fill: yellow)),
    // Red edge visible only on slides 3-4
    between(3, 4, edge((1, 0), (2, 0), "-|>", stroke: red, label: [3-4])),
    // Green edge appears from slide 4
    step(4),
    node((2, 0), [Near], radius: 2em, fill: green.lighten(50%)),
    // Final edge from slide 5
    step(5),
    edge((2, 0), (3, 0), "-|>", stroke: purple, label: [final]),
  )
]
```

I also exposed step, because it can be hard to keep track of the count of preceding `pause`s and mix it with the `between` calls.

PR #300, opened 2026-03-11
Feature: Implicitly allow fn-wrapper based animation functions via reducer

zral0kh:
Having to go through writing the callback style slides for using `uncover`/`only` and similar in cetz/fletcher is a hassle. Especially because of the whole new self creation and that we have to count our subslides manually.

#247 wanted to allow  such commands in fletcher directly, but used weird new names.
We should, for consistency, simply use the existing ones.

Changes:
 - create the reducer self automatically, 
 - fixed some minor inconsistencies and 
 - also allow waypoints in these functions now.

Added some small tests for cetz/fletcher to test these fn-wrapper reducer-based calls.

For Fletcher the syntax becomes a lot simpler, for CeTZ we need to call functions inside arrays, just like we normally did with pause:
`(uncover("2-", rect((0,0),(1,1))), )"

This should do all that #247 wanted, but in a consistent way.

### DIFF
--- a/docs/en/integration/cetz.md
+++ b/docs/en/integration/cetz.md
@@ -77,13 +77,13 @@ In fact, we can also use `only` and `uncover` within CeTZ, but it requires a bit
 #show: simple-theme.with(aspect-ratio: "16-9")
 
 #slide(repeat: 3, self => [
+  #let (uncover, only, alternatives) = utils.methods(self)
 
   Cetz in Touying in subslide #self.subslide:
 
   #cetz.canvas({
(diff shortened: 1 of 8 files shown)


zral0kh:
perhaps requires some additional documentation that we now allow this as well.

zral0kh:
had some problems with implicit waypoints and also alternatives

zral0kh:
should be ready right now, please test it yourself a bit @OrangeX4

OrangeX4 (approved):

PR #301, opened 2026-03-12
Feature: Quality of Life -- auto, "h"-here string and inverse function for string subslide-numbers and waypoints

zral0kh:
This PR packages a lot of small quality of life additions for writing touying slides, especially surrounding not having to count subslides anymore. 

## `auto` for `uncover`, `only`, `effect`
We now allow `uncover(auto)[Text]` and similar for `only` and `effect`, which uses the subslide index at the position in the content tree.

## Pseudo string number "h" and derivations
We allow the pseudo string subslide-number "h" and derivations thereof.
"h" means here, and behaves just like `auto`.
Possible options are: "h", "h-", "-h"

This means you can write `effect(text.with(color:red), "h-")[Important text]` now which will be shown red once the subslide is reached given the other things outside.

This brings the functions on par with normal writing of text via pause since we no longer have to care about counting the subslide index ourselves.

## String Range Inversion via "!"
We allow inversions of those subslide number strings via the "!" pseudo function.
Simply write 
- "!h"/"!2" - not here, or not at index 2
- "!-3" - not(until 3)==after index 3 (similar for "!3-")
- "!2-4" - less than 2 or larger than 4, outside of that range.

The main idea is that this allows us to invert e.g, an only, without having to specify all other subslides.

> Note: The inverse does not inflate the subslide count. So you can write "!4" even if your slide only has 3 subslides. This would just evaluate to "1,2,3".

## Waypoint Inversion via `not-wp`
This behaves exactly like the pseudo function "!", but may be a bit counterintuitive in some cases.
Remember that waypoints capture the entire range of subslides until a new waypoint is declared.
Given
```typst
#slide[
A #waypoint(<B>) B1 #pause B2 #waypoint(<C>) C

#only(not-wp(<B>))[Not at B]
]
```
'Not at B' will be shown when A and C get uncovered. Neither when B1 nor when B2 get uncovered, because both fall into the range of the waypoint `<B>`. This means only at subslides 1 and 4 and not at 2 and 3.

### DIFF
--- a/docs/en/tutorials/dynamic/waypoints.md
+++ b/docs/en/tutorials/dynamic/waypoints.md
@@ -51,6 +51,10 @@ The implicit waypoint is only registered once per label (the first occurrence wi
 ]
 ```
 
+This produces 4 subslides: items appear on 2, 3, 4 (the implicit `<list>` waypoint advances to subslide 2), and `<done>` fires on subslide 5. 
+
+Note: Waypoints capture all subslides into their range until a new waypoint follows.
+
 ## Non-advancing Waypoints
 
 By default, waypoints advance the subslide counter. Use `advance: false` on explicit waypoints to mark a position without creating a new subslide:
@@ -61,10 +65,6 @@ By default, waypoints advance the subslide counter. Use `advance: false` on expl
   Content at the current position.
 ]
 ```
-
-This produces 4 subslides: items appear on 2, 3, 4 (the implicit `<list>` waypoint advances to subslide 2), and `<done>` fires on subslide 5. 
-
-Note: Waypoints capture all subslides into their range until a new waypoint follows.
 
 ## Waypoint Markers
 
--- a/docs/zh/tutorials/dynamic/waypoints.md
+++ b/docs/zh/tutorials/dynamic/waypoints.md
@@ -55,9 +55,20 @@ sidebar_position: 7
 
 注意：路标会捕获其后的所有 subslides 直到下一个路标出现。
 
+## 非前向路标
+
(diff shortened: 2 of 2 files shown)


zral0kh:
This still requires documentation and potentially new tests. 

I did test all those things, but since waypoints introduced so many new tests and the rest of the code is not that well covered, i thought it better to not introduce many more tests rn, since that increases test running times a lot.


@OrangeX4 please say whether/where to put such.
And also if you are happy with these things or sth should change.

OrangeX4:
I think the documentation could be put in `docs/en/tutorials/dynamic/complex.md`. For testing, I don't think we need to worry too much about the size and running time, as long as the test is sufficient, you can do as you think.

OrangeX4:
You can set width 80 for typstyle (default config for typstyle but not tinymist):

image

docs/en/tutorials/dynamic/waypoints.md
```diff
@@ -51,13 +51,24 @@ The implicit waypoint is only registered once per label (the first occurrence wi
 ]
 ```
 
+## Non-advancing Waypoints
+
+By default, waypoints advance the subslide counter. Use `advance: false` on explicit waypoints to mark a position without creating a new subslide:
```
OrangeX4, line 60: Is this part here correct? Below: `This produces 4 subslides: items appear on 2, 3, 4`

zral0kh:
i had moved the non-advance waypoint up so i can use it below and misplaced it slightly.

Should be fixed now also the chinese doc.

Issue #136, opened 2024-12-11
`align(horizon)` with `#pause` in between bullet points results in slightly shifted slides


t1mlange:
Hi,

consider following reproducer:
```typst
#import "@preview/touying:0.5.3": *
#import themes.metropolis: *
#import themes.simple: *

#show: simple-theme.with(
  aspect-ratio: "16-9",
  config-info(
    date: datetime.today(),
  ),
)

== Title

#slide()[
#set align(horizon)
- A
#pause
- B
]
```

results in following two pages:
![image](https://github.com/user-attachments/assets/1d81e038-5153-4c7d-bd5f-b1806b336532)

As you can see, the first bullet point is not correctly aligned. Do you have any tip on how to fix this?

enklht:
The issue seems to be that the `#pause` directive splits a list into two parts. Here is a workaround I came up with:

```typst
#import "@preview/touying:0.5.3": *
#import themes.metropolis: *
#import themes.simple: *

#show: simple-theme.with(
  aspect-ratio: "16-9",
  config-info(
    date: datetime.today(),
  ),
)

== Title

#slide()[
#set align(horizon)
#set par(spacing: 0.65em)
- A
#pause
- B
]
```

This works because the size of a space between list items inherits the paragraph's leading parameter, which is set to 0.65em in this case.

In the future, it would be beneficial if the `#pause` could respect the original document structure.

enklht:
Sorry, I initially thought it was a problem with `pause`, but I found that it was actually due to `hide`. We cannot avoid this issue in touying, at least with normal methods.

If you use another cover function, such as the newly implemented `alpha-changing-cover.with(alpha: 0%)` (not included in the current version), the problem doesn’t occur. However, you cannot hide markers (numbers) in lists (enums) using this method.

t1mlange:
I can't follow you all the way, but setting the par spacing to the same value as the list spacing works fine with a slide set only containing figures and enums/lists.

OrangeX4:
More discussion: https://discord.com/channels/1054443721975922748/1280244268769873950/1319663908688232482

Andrew15-5:
```typ
#import "@preview/touying:0.5.5": *

// Not affected:
// #show: themes.simple.simple-theme
// #show: themes.university.university-theme
// #show: themes.aqua.aqua-theme
// #show: themes.dewdrop.dewdrop-theme
// #show: themes.default.default-theme

// Affected:
// #show: themes.stargazer.stargazer-theme
[...]

Issue #225, opened 2025-10-16
List items with `#pause` are moving with metropolis or stargazer theme


Orbion-J:
Hello,

I noticed a strange behaviour using lists with the metropolis theme. In the minimal example below, the first item is not at the same position in each page. It is slightly higher in the first one.

```typst
#import "@preview/touying:0.6.1": *
#import themes.metropolis: *

#show: metropolis-theme.with(
  aspect-ratio: "16-9",
)

==

- #lorem(5) #pause
- #lorem(5)
```

I could reproduce this behaviour with the stargazer theme but not with the other ones.

The slightly more advanced example below have items that move up and down on each page. 
```typst
#import "@preview/touying:0.6.1": *
#import themes.metropolis: *

#show: metropolis-theme.with(
  aspect-ratio: "16-9",
)

==

- #lorem(5) #pause
- #lorem(5)

#lorem(10) #pause
- #lorem(5) #pause
- #lorem(5) #pause
  - #lorem(5) #pause
  - #lorem(5)
```

tfardet:
This is also happening with the university theme, there is a 0.55em vertical drift per pause

lennart-k:
Having the same issue although it's only the last list item causing the layout shift.

I've made a small example which shows the body size in red.
It seems like as long as list items are hidden the list is split into two which have a margin between each other.
As we uncover the last item the two lists are merged into one causing our layout to be smaller by the margin between two lists.
```typ
#import "@preview/touying:0.6.1": *
#import themes.simple: *
#show: simple-theme

== Section

#align(horizon, block(fill: red)[
- Okay wow
#pause
- Nice
#pause
- Last list item causes layout shift
#pause
Text is fine :)
])
```

PR #303, opened 2026-03-13
Fix cover spacing issues surrounding lists: clean

zral0kh:
When lists are broken up into pieces due to the default hide function we get spacing changes bc typst does not recognize the list continuing anymore. 

The problem is reported in #136 and #225.

The fix applies to all three: `list`, `enum`, `terms`, did not test yet whether this causes problems with other `cover`-functions.

In short form this fix tries to recognize whether a `list.item` has a previous `list.item` and if so we apply a `block(spacing: par.leading)`, obv. a bit more twisted than that,

This still breaks when the user locally uses
```typst
show list.where(tight: true): magic.nontight
``` 
but we cannot fix that, bc tightness is set on the list, not the items.
(maybe with a large rewrite, but not worth it)

Rectangle based cover methods are also broken, but not sure how to fix that, the `box` used is the problem.
Color and Alpha Cover Modes verified working.

Fixes: https://github.com/touying-typ/touying/issues/136 and https://github.com/touying-typ/touying/issues/225.

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -3195,11 +3195,11 @@
       let covered = cover-fn(items.sum())
       //decrease below spacing for rect cover functions
       // if type(cover-fn) == function and (
-      //   cover-fn==utils.cover-with-rect or 
+      //   cover-fn==utils.cover-with-rect or
       //   cover-fn==utils.semi-transparent-cover
       // ){
       //   covered // does not fix it, but does not hurt: problem stems from box itself causing later content to be shifted? idk
-      // }else 
+      // }else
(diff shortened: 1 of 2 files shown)


zral0kh:
If someone has a good idea how to fix the rectangle based covers, please fix, otherwise i support merging this and opening an issue for the explicit problem such that it is reported.

OrangeX4:
Thanks for this PR, I think it would be better to merge this for now.

OrangeX4 (approved):

PR #304, opened 2026-03-13
Waypoints feature: start param and Waypoints in `handout-subslides`

zral0kh:
1. Waypoints now allow for a param `start`, which accepts single integers and other waypoint labels.

E.g. you now can do sth like:
```typst
== Start Int/Label Backward

// A waypoint can jump backward to an earlier subslide.

#waypoint(<bk-a>, advance: false)
Phase A.

#pause
Subslide 2.

#pause
Subslide 3.

#waypoint(<bk-back>, start: 1)
Back to 1. (with A)

#pause
Now at 2.

#waypoint(<bk-again>, start: <bk-a>)
1. #pause 2. #pause 3.

#only(<bk-back>)[During first back (1,2).]
#only(get-first(<bk-back>))[Back-jump lands on subslide 1.]
```

However you would need that idk, but it may make some things easier to write, e.g. descriptions for animated graphics or sth.

It basically is a jump which allows waypoints, but now waypoint ranges can also truly overlap with each other.

2. `handout-subslides` can now be parsed a waypoint label or a waypoint marker. If a range results from this, only the first subslide is chosen. E.g. putting in a waypoint label will result in a handout slide replaying the slide at the subslide index where the waypoint is defined, not where its reach ends. (completes #305)

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -5344,48 +5344,9 @@
         ..bodies,
       )
       header = page-preamble(self) + header
-      set page(
-        ..(self.page + page-extra-args + (header: header, footer: footer)),
-      )
-      body-transform(setting-fn(
-        subslide-preamble(self) + composer-with-side-by-side(..conts),
-      ))
-    } else {
-      //negative indices in string not defined/supported, and they can even have ! for inversion.
-      let handout-subslides = _parse-negative-subslide-indices(
-        self,
-        handout-subslides,
-      )
-      // Render only the subslides that match handout-subslides
-      let handout-subslide-indices = range(1, repeat + 1).filter(
-        i => utils.check-visible(i, handout-subslides),
-      )
-      // Fall back to the last subslide if none match
-      if handout-subslide-indices.len() == 0 {
-        handout-subslide-indices = (repeat,)
-      }
-      let result = ()
-      for (pos, i) in handout-subslide-indices.enumerate() {
-        let is-first = pos == 0
-        let is-last = pos == handout-subslide-indices.len() - 1
-        let subslide-self = self
-        subslide-self.subslide = i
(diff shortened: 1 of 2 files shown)


OrangeX4:
Each change to the waypoint feature can alter over a hundred images in the waypoint test. For example, imagine adding a new slide; all the total page numbers will change. Consider removing page number display from the slides. Ideally, divide the waypoint tests into several smaller tests and place them in a folder like tests/waypoint.

zral0kh:
true, sorry should be last time hopefully

zral0kh:
i am thinking of checking the behaviour of hierarchical waypoints in the `start` field before marking as ready, 

and allowing waypoint markers inside `only`/`uncover` inside touying-reducer, which currently does not work.

With both those things done i think waypoints are finally complete.

OrangeX4:
I'm thinking of releasing 0.6.4 to fix several bugs that are currently encountered in 0.6.3, so the merged waypoints feature should also be released at the same time. Do you think this PR still needs to be merged before 0.6.4? It can also be moved to versions after 0.6.4 :-)

zral0kh:
will do the remaining things for waypoints now, so it can be in 0.6.4 (allow allow waypoint markers in reducer)

btw i think we could even up the version to 0.7.0 as some breaking changes happened, especially with how show-rules are now handled and how config can now be set.
and the whole waypoint feature should also be big enough for that.

zral0kh:
actually it turns out the waypoint prepass already properly traverses the reducer and thus waypoint markers work as well, the function written to the subslides (which gets resolved to the correct subslide index in utils) is correct even if the touying-fn-wrapper happens to be inside touying-reducer.

OrangeX4 (approved):

OrangeX4:
Thanks!

PR #313, opened 2026-03-17
refactor: simplify extraction of hidden element functions

Andrew15-5:
I've noticed https://github.com/touying-typ/touying/issues/136 got fixed, then quickly looked at [the changes](https://github.com/touying-typ/touying/pull/303/changes). Nothing caught my eyes except for [the very first part](https://github.com/touying-typ/touying/blob/30930c30f367e07621f4afc132205b53e6208567/src/core.typ#L3256) with extraction of the hidden `space` element (?) function.

I believe that I already know the most simplified extraction of all hidden functions (I'm not exactly sure who first authored each solution though) that I've been using for years now. They still work with no changes (between versions). I can't remember in which projects did I use them other than the latest, which is the IJIMAI template. [Here are all of them together](https://github.com/pammacdotnet/IJIMAI/blob/587e372ce0ce7c67b0e2ba2c43e7337c29ccddf1/ijimai.typ?plain=1#L253-L257).

In this PR, I've used those solutions to replace bulkier ones in this project. More precisely: `sequence`, `styled`, `space`.

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -3251,9 +3251,6 @@
     let result = ()
     let hidden-parts = ()
 
-    // Reference element for detecting space nodes in the content tree.
-    let _space-func = [ ].func()
-
     // Helper: is this content element a list/enum/terms item?
     let _is-list-item(it) = (
       type(it) == content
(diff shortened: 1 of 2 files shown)


Andrew15-5:
> i renamed the space function to be named similar to the other, please approve and then we can def merge

I don't understand this. "Similar" to what? I didn't change any names. The `_space-func` was created (in https://github.com/touying-typ/touying/pull/313), not renamed. What exactly do you want me to approve?

zral0kh:
ah sry i am not allowed to change your stuff directly, i made a pr.

but you can also just add it to utils as we might need it somewhere again anyway and that keeps it clean as well

Andrew15-5:
> add it to utils as we might need it somewhere again anyway and that keeps it clean as well

Done.

OrangeX4:
Thanks.

Issue #208, opened 2025-08-12
Cannot use alert inside only


Beiri22:
Hi, I am wondering why I cannot use alert() inside only()... like 

```
#only("1-2,4")[
  *Some list*
  - #alert(2)[Item 1]
  - #alert(4)[Item 2]
]

#only(3)[
    #image("imgs/A.png")
]

#only(5)[
  #image("imgs/B.png")
]
```

I get  panicked with: "Unsupported mark `touying-fn-wrapper` at page 43. You can't use it inside some functions like `context`. You may want to use the callback-style `uncover` function instead."

Beiri22:
The same stops me from checking inside only() if we are in handout mode, like:
```
#only(4)[
  // Nur, damit keine Markierung im Handout ist.
  #touying-fn-wrapper((self: none) => [
    if #self.handout [...]
  ])
]
```

Why I need this: I am drawing onto the slide as a form of highlight. I want to show this on slides, but not on handout.

zral0kh:
first off: alert was never meant to be called with a subslide index.

secondly you can use `#handout-only` to show handout only content, see the docs for details.
and in some future version we will likely also allow `#presentation-only`

still both things are currently not allowed since we don't traverse the internal content of fn-wrappers at the moment. This might change in the future. `alert` and `handout-only` are fn-wrappers themselves and detecting them would need us going into the fn-wrappers inner content. Which would allow a whole set of nesting etc.

You problem will likely be fixed by `presentation-only` and for `alert` since it will (via #316 ) be based on a raw fn wrapper instead we could change that without allowing deep nesting.

I think there probably is some sketchy way to get it working even now, but not sure.

zral0kh:
@OrangeX4 can be closed now i think

PR #316, opened 2026-03-19
Fix: Alert not delayed

### LINKED ISSUE #311: `alert` content can't be hidden in 0.6.3
`alert` content seems unaffected by `#pause` since 0.6.3, e.g. in

```typst
#import "@preview/touying:0.6.3": *
#import themes.simple: *

= Title

== First Slide

Hello, Touying!

#pause

Hello, #alert[Typst!]
```

`#alert[Typst!]` is visible before the `Hello` preceding it.

zral0kh:
Introduces a new `touying-fn-wrapper-raw` that still passes self to functions, thus makes them settable on slides and via config, but no longer requires subslides/repetitions to be passed to it, new branch on the split subslides function in `core.typ`.

Replace `alert`'s `touying-fn-wrapper` with the raw variant.

Fixes #311 

Now also allows `touying-fn-wrapper-raw`, e.g. `alert` inside `touying-fn-wrapper` content, e.g. in `uncover`.


zral0kh:
perhaps this still needs some documentation change?
and we might want to add now allowing fn-wrapper-raw inside of normal fn-wrapper's content to fix #208 or rather what it wanted.

zral0kh:
Created an issue for remembering the problems around nesting `touying-fn-wrapper-raw` inside `touying-fn-wrapper. With an addition to the docs, explaining current shortcomings/how one may use and not use `alert`, we should be able to merge.

zral0kh:
now also fixes that issue:

#325

OrangeX4:
Thanks

PR #317, opened 2026-03-20
Fix #312 ghost slides with show rules. fix #318 + proper consistent handling of show rules and `defer` keyword

### LINKED ISSUE #312: Empty slide after all show-rules in stargazer-theme 0.6.3
#### Description
Since `touying:0.6.3`, an empty slide will be inserted before the appendix if we put anything invisible next to ~the appendix show-rule~. **Edit:** This really affects all show-rules😕, see https://github.com/touying-typ/touying/issues/312#issuecomment-4082599504, thanks @zral0kh.
- like this:
  ```typst
  // content slides
  #show: appendix
  [#metadata((kind: "touying-appendix"))<touying-metadata>]
  #counter(heading).update(0)
  // appendix slides
  ```
- or like this:
  ```typst
  // content slides
  [#metadata((kind: "touying-appendix"))<touying-metadata>]
  #counter(heading).update(0)
  #show: appendix
  // appendix slides
  ```
Note that even if we remove either of the two lines `[#metadata(...)<...>]` or `#counter(heading).update(0)`, the bug still occurs. Personally, this is annoying and makes it impossible to impose some rules for the appendix of my custom template.

I have only encountered this for the `stargazer-theme`, but other themes such as `dewdrop-theme` seem to work fine _(need testing for others)_. It was also not a thing in previous Touying versions.

#### Minimum working example
```typst
#import "@preview/touying:0.6.3": *
#import themes.stargazer: *

#show: stargazer-theme.with(
  aspect-ratio: "16-9",
  config-info(title: [Title]),
)

== Frame Title
There should be only 1 contents slide.

#show: appendix
#counter(heading).update(0)  // This line

== Backup Slide
Appendix.
```

##### Expected behaviors
Two slides produced.

##### Observed behaviors
Three slides produced.

### LINKED ISSUE #318: An empty last slide is not counted with `#show: appendix`
## Minimal source to reproduce
```typst
#import "@preview/touying:0.6.3": *
#import themes.simple: *

#show: simple-theme.with(
  aspect-ratio: "16-9",
  footer: [],
  footer-progress: false,
  config-info(
    title: [Title],
    author: [Author],
  ),
)

== 1

== 2

#show: appendix

== Appendix
```
## Output
Image

The issue goes away when you either remove `#show: appendix` or add some content to the last slide.
```
== 2
a
```

zral0kh:
As explained in #312 and #315 the behaviour stems from a deep underlying problem of how style show rules and especially config show rules were processed in `split-content-into-slides`.

The old behaviour was:
- when you encounter a style node, go into and split if necessary. If not recombine everything with the old slide
- when you encounter a config change split into new slide.

The new behaviour is:
- when you encounter style or config change split and recombine all parts from the content tree until explicit other split is declared, e.g. heading or pagebreak or `slide` function call: for style split the style change, for config update split the config change. Config changes take affect only after a subslide increment ofc.
- rest of the content after the explicit slide split is affected from the changes via the show rule
- appendix is reimplemented via `defer` keyword, which puts that and all following calls into a `leading-preamble` which gets called on the next slides `slide-preamble` call, before anything else, even before the normal `self.preamble` is processed,

The `defer` keyword is thus also interesting to set the next slides config without using the explicit function `slide`, this includes setting the normal `preamble` function.

Also now fixes #318, where the appendix show rule would apply to the current slide if it is empty, not the next one.
For deferred config show rules, we now properly flush the headings, even if no content was there.

---

This PR further includes an update to the `cover` test's ref which now uses the new functionality of config changes not splitting slides anymore, but being applied on the next subslide. And a `features/show-rules` test, which tests all this behaviour for edge cases.


zral0kh:
this definitely needs some documentation before merging

OrangeX4:
Thanks!

PR #320, opened 2026-03-20
fix: function `alternatives-match` takes into account parameter `stretch`

Cemoixerestre:
Previously, this parameter was ignored, there was no stretching.

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -1430,7 +1430,7 @@
     },
     subslides-contents,
     position: position,
-    stretch: false,
+    stretch: stretch,
   )
 }
 


Cemoixerestre:
There was a discrepancy between the function `alternatives` and `alternatives-match`. Now, the following code produces the same result for both:
```typst
#import "lib.typ": *
#import themes.simple: *
#import "@preview/codly:1.3.0": *

#alternatives-match(("1": "a", "2": "bcdefghikl"), stretch: true) with `alternatives-match`

#alternatives("a", "bcdefghikl", stretch: true) with `alternatives`
```

OrangeX4:
Thanks!

Issue #321, opened 2026-03-21
Page margin `x`/`y` does not override the previously set 4 sides (`top`/`right`/`bottom`/`left`)


Andrew15-5:
The problem from [the forum](https://forum.typst.app/t/touying-config-page-margin-y-values-are-not-working/8301?u=andrew), where `y` values does nothing as [explained by me](https://forum.typst.app/t/touying-config-page-margin-y-values-are-not-working/8301/3?u=andrew):

```typ
#import "@preview/touying:0.6.3": *
#import themes.university: *
#show: university-theme.with(config-page(margin: (x: 2em, y: 4em)))

== Test
#lorem(20)
```

PR #322, opened 2026-03-21
fix: correctly handle page margin merge/precedence

Andrew15-5:
Fixes https://github.com/touying-typ/touying/issues/321.

I quickly wrote the fix with some comments, but I'm not 100% I thought of all (edge) cases. But I think the main issue is now fixed.

Does the project use American English like Typst or British English? It isn't stated anywhere I looked in.

I assume there are no `config-*` with stuff like [`rect.stroke`](https://typst.app/docs/reference/visualize/rect/#parameters-stroke), which is slightly different from `page.margin`.

I'm pretty sure it's handled correctly in `set page` rules, so probably can just yonk the Rust code and RiiT. That is, if it has the special margin/precedence handling (surely) unlike plain "dictionaries".

### DIFF
--- a/src/utils.typ
+++ b/src/utils.typ
@@ -1,5 +1,35 @@
 #import "pdfpc.typ"
 
+/// Add page margin dictionary to another page margin dictionary.
+///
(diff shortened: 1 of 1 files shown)


OrangeX4 (approved):

OrangeX4:
Thanks!

PR #328, opened 2026-03-27
Fix `cover-with-rect` breaking long lines of text when partially hidden and fallback functions for color/alpha cover

zral0kh:
It took a while but i finally thought of a solution on how we can get cover-with-rect to not break text into a new line: use the `strike` function which basically already is a rect and we simply set the stroke width to the lineheight+ some padding.

Falls back to the rectangle cover for inline math mode (strike does not work there) and block content.

I also "fixed" it not hiding block content properly by just setting the width to the line width. This works well for semi-transparent, but is a bit ugly for pure rectangles, but not sure how we can get the local alignment of aligned/placed blocks.

And since this now is a good alternative to the normal hide (finally), i allowed setting a fallback function instead of always using the typst `hide`, still defaults to that one though.

And i changed the cover test to test for some more edge cases.


zral0kh:
btw this also fixes rect cover now properly doing the cover-spacing around lists :)

zral0kh:
some minor fixes left for correct sizing of strike cover on large/small text

zral0kh:
nah can't get font size changes with par show , e.g. `show par: set text(size: 2em)` to work. 
it seems text is only separated into pars in the final layout construction and thus either i don't get the local text size even with context or i need to wrap in a layout call which breaks the whole idea of using strike (by causing a linebreak again if text is too long)
other element text sizes, even if changed via show that is block: e.g. `skew` is measured correctly

maybe we can live with that? idk 
then the user will need to set inline to false which will just render everything as block regardless. or we allow them to manually specify a text size for inline text which they can use when the normal functionality does not meet their requirements.

zral0kh:
Anyway i pulled it off somehow.
Turns out we can just wrap the inline content in `par` before measuring which gives us the right size.

And then it turns out that blocks that consist only of text inherit the increasing spacing size depended on their scaled text, e.g. via `show skew: set text(2em)` that will also double the leading and trailing padding for blocks, but since we cannot wrap in a block bc that would destroy alignment, we use 2 pad functions instead.

zral0kh:
its a bit complicated now, but that is what it takes i guess..,

zral0kh:
there still might be some mini issues about `align`, see e.g.
```typst
#show par: set text(2em)
#align(center)[hello]
#pad(top:1em, align(center)[hello])
```

the aligned text in align will get the increased par size, but the padded aligned text won't and will be normal size.
but not sure how we can fix that, as align cannot be reconstructed bc the alignment is not a property on content.

this `#pad(top:1em, align(center)[#par[hello]])` would fix the sizing but as explained above we cannot do that. so we just have to note somewhere that this is a limitation and this is the way to fix it

zral0kh:
got align to work via some weird magic.

last remaining thing i couldn't solve:

text only blocks with larger text due to styles will also have enlarged spacing. but their spacing collapses partially when they follow other large text only blocks, we cannot detect that, maybe we would have to do that outside the utils and just like with with the list cover spacing then also track text blocks following each other, but that should happen rarely. in 99.99% of cases using an alpha cover and the semi-transparent as fallback cover for non-text blocks should be enough.

OrangeX4:
Thanks, I think that's enough for now

OrangeX4:
0.7.0 has already been released, thank you for the fixes and features you added to touying. You are now touying's second contributor. I recently posted 0.6.3 in the discord showcase. If you want, you can also post an update announcement for 0.7.0 on discord, or you can wait until more updates are added. This might help others recognize the many new features you've contributed to touying :-)

PR #332, opened 2026-03-30
fix fit-to-height and size-to-pt and allow text reflow

### LINKED ISSUE #109: `utils.fit-to-height` no longer working on typst 0.12

zral0kh:
it seems the ability to measure fraction stuff via v,h elements is no longer there, thus we extract the float instead and use that.

fixes #109

### DIFF
--- a/src/utils.typ
+++ b/src/utils.typ
@@ -1,5 +1,4 @@
 #import "pdfpc.typ"
-#import "magic.typ"
 /// Add page margin dictionary to another page margin dictionary.
 ///
 /// Example: `add-page-margin-dicts((top: 1cm, x: 2cm), (y: 3em))` returns `(x: 2cm, y: 3em)`
@@ -1125,7 +1124,7 @@
 ///
 /// - body (content): The content to fit. If two positional arguments are given, this will be width instead.
 ///
-/// - args (arguments): For convenience and compatibility with older versions, passing in width as a positional argument is still supported. If two positional arguments are given, the first one is the width and the second one is the body. 
+/// - args (arguments): For convenience and compatibility with older versions, passing in width as a positional argument is still supported. If two positional arguments are given, the first one is the width and the second one is the body.
 ///
 /// -> content
 #let fit-to-width(width: 1fr, grow: true, shrink: true, body, ..args) = {
@@ -1134,9 +1133,6 @@
     message: "Only two positional arguments allowed, which will be interpreted as width and body.",
   )
   if args.pos().len() == 1 {
-    magic.warning(
-      "Passing 'width' as a positional argument is deprecated, please use the named argument instead.",
-    )
     width = body
     body = args.pos().at(0)
   }


OrangeX4:
I think this issue hasn't been fixed yet? For this example

```typ
#import "/lib.typ": *
#import themes.simple: *

= Title

== First Slide

#lorem(50)

#utils.fit-to-height(1fr)[
  #lorem(50)
]
```

zral0kh:
what do you expect to happen?
that the `fit-to-height` block does not break the page and instead adjust to fill the remaining space? 

If so i would like to suggest we also set width to `auto` by default, which tries to use the full page/container width after scaling,

zral0kh:
see how the example now flows.

Now i know that is not perfect, but since we cannot access the layout algorithm this is almost the best we can do.
We could make some more sqrt refinement iterations given the remaining space, but not even sure whether typst allows that.

Maybe what would be easy to do is have one final simple refinement step: measure the content, width will be smaller than container-size, then scale the boxed-content to fit that width. Since it should be pretty close already it and linebreaks are sort of discrete and quantized (only appear in certain weird steps), it does not make sense to reflow often anyway. doing it once and then rescaling should be enough.

zral0kh:
so i got it to a point where i don't know how to improve it rn.

see e.g.
grafik

the text cannot flow into 4 lines without the height not matching: 4 lines occupy about 78% of space of 5 lines, meaning to scale height back to the 1fr we would scale by `1.28` which then overshoots in width direction.

There is no good solution for this i fear. 
but its still a lot better than the primitive implementation.

The only question is what the preferred behaviour is when `width: auto, reflow:true`:
-  should it fill the width and ignore the last bit of height since we cannot do better, 
- or do we really want to fill the height like a maniac.

we could also give the user some field `force-height` which manages that decision

zral0kh:
well thats what i did rn.

OrangeX4:
I think the current result is already very great, this is the expected capability, to be able to occupy the remaining page space with for contents like `#utils.fit-to-height(1fr)[#rect[]]`. Thank you for the implementation. However, we have three small suggestions:

1. You can add a doc comment to `force-height` argument
2. Maybe `reflow` could be set as the default to `true`, as it might improve user experience
3. We could consider changing `height` to a named parameter with a default value of `1fr`, and then adding `..args` to maintain compatibility. Then, if `args.pos().len() == 1`, we would only extract it as `height`; otherwise, we would default to the named parameter `height`. The same applies to `fit-to-width`. Because I've observed that most of the time users just need to input `1fr`, having it as the default value would be much better.

zral0kh:
if i set `reflow` to true, we have waste compute on every call where its not text, maybe we can check if the content has text beforehand? or we just do it

OrangeX4:
We can indeed do a simple mechanism, if reflow is set to auto, then we will simply see if the content is text, if so, turn it on, if not, turn it off.

zral0kh:
done

src/utils.typ
```diff
@@ -843,21 +882,44 @@
...
+    let layout-content(
+      width: auto,
+      prescale-width: none,
+      grow: true,
+      shrink: true,
+      height,
+      body,
+    ) = layout(container-size => {
       let available-height = _size-to-pt(height, container-size.height)
       // Provide a sensible initial width, which will define initial scale parameters.
```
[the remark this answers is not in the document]
zral0kh: true

zral0kh:
perhaps we can later think of an improved reflow mechanism that tries to maximize the covered area by the text.
this should then actually make it use up all the space. but not sure how we can achieve that

zral0kh:
@OrangeX4 and perhaps we should think about how we handle deprecation. let's discuss in some discussion forum perhaps.

src/utils.typ
```diff
@@ -949,18 +1120,32 @@
...
 /// -> content
-#let fit-to-width(grow: true, shrink: true, width, content) = {
+#let fit-to-width(width: 1fr, grow: true, shrink: true, body, ..args) = {
+  assert(
+    args.pos().len() <= 1,
+    message: "Only two positional arguments allowed, which will be interpreted as width and body.",
+  )
+  if args.pos().len() == 1 {
+    magic.warning(
+      "Passing 'width' as a positional argument is deprecated, please use the named argument instead.",
```
OrangeX4: I think both of these parameter passing methods are feasible, and there's no need to prompt deprecated? This usage is quite common, for example in typst's built-in `set text(12pt)` and `set text(size: 12pt)`, or in the theorem package's `#theorem[Title][Body]` and `#theorem(title: "Title")[Body]`.

tests/features/fit-to/test.typ
```diff
@@ -0,0 +1,32 @@
+#import "/lib.typ": *
+#import themes.simple: *
+
+#show: simple-theme
+
+= Fit To Test
```
OrangeX4: Perhaps we can remove this level 1 heading and then change the subsequent level 2 and level 3 headings into level 1 and level 2 headings respectively? This way, we can avoid using `---` and creating empty pages.

src/utils.typ
zral0kh: okay

src/utils.typ
OrangeX4: here also

OrangeX4: and typstyle

zral0kh: yeah sorry, really thought i did that

Issue #331, opened 2026-03-30
Feature: Ability to access a slide's config (#210)


zral0kh:
In the discussion #210, a user wants an easier way to access the config.

We should expose some direct way for this: e.g.
`touying-get-config` similar to `touying-set-config`, but this one is parsed and rendered to just the value at the location requested.

it should work sth like

```typst
#touying-get-config("info.date")
```

which gets evaluated to `self.at("info", default: (:)).at("date", none)` if we don't want it to throw an error or we check the path upfront.

OrangeX4:
I think this is a nice feature, but I'm thinking about whether we should simplify these usages, such as removing the `touying` prefix. Or we could look for other more convenient syntax.

PR #333, opened 2026-03-30
feature: `touying-get-config` function

zral0kh:
a function that retrieves the config value for a given key as content, uses the slide self.
was first mentioned here: #210 and it completes #331.

you may use it like
```typst
#touying-get-config("info.author", default: [My Name])
//or also
#touying-get-config().info.author
```

~you cannot retrieve the complete self, not sure if that is wanted.~

Requires some documentation i guess.


src/core.typ
```diff
@@ -822,6 +833,34 @@
...
+          config-value = config-value.at(key)
+        } else {
+          config-value = default
+          break
+        }
+      }
+      if type(config-value) != content {
+        // Convert non-content config values to strings for display, so that
+        // touying-get-config can be used in any context without type errors.
+        config-value = repr(config-value)
```
OrangeX4: Why is `repr` used here? I think `config-value = [#config-value]` might be a better solution

src/core.typ
```diff
@@ -822,6 +833,34 @@
...
+      }
+      slide-parts.push(config-value)
     } else if utils.is-styled(child) {
       // When absorbing leading preamble and no heading seen yet, recurse into
       // the styled node with absorb-leading-preamble: true. The set/show rules
```
OrangeX4, line 829: Should we default to throwing an error instead of using the default value? Generally, we want to stop early with errors to prevent users from ignoring them, while the default is explicitly set by the user?

src/core.typ
```diff
@@ -822,6 +833,34 @@
...
         }
       }
+    } else if utils.is-kind(child, "touying-get-config") {
+      assert(
+        type(child.value.key) == str,
+        message: "The key argument of touying-get-config must be a string. e.g. \"info.date\"",
+      )
+      assert(
+        child.value.key.starts-with(
+          regex("[a-zA-Z0-9\\-]+(\\.[a-zA-Z0-9\\-]+)*$"),
```
OrangeX4: We might need to think about this, as we cannot assume users will always input in this format. For example, users might enter underscores, other languages like CJK, or other alphabets. If we can't determine it well, it's also acceptable to simply remove the assert.

src/core.typ
```diff
@@ -34,6 +34,17 @@
   body: body,
 ))<touying-temporary-mark>]
 
+/// Gets the current config value for the given key at the slide where this is called. This is useful for implementing features that need to read config values. The result is only available during `context` resolution, thus you should not compute with it. Just print it.
+///
+/// - key (str): The key of the configuration to retrieve.
+/// - default (any): The default value to return if the key is not found.
+/// -> any
+#let touying-get-config(key, default: none) = [#metadata((
```
OrangeX4: I've always wanted to simplify the names `touying-set-config` and `touying-get-config` (of course, their old names can be kept to ensure compatibility, just with aliases added). However, I haven't had a great idea—perhaps just `set-config` and `get-config` could be simpler?

docs/en/faq.md
```diff
@@ -840,6 +840,41 @@ Use `touying-set-config` around the content you want to change:
...
 ```
 
+
+### How do I access global or slide config information?
+
+Use `touying-get-config` at the position you want to know the config. The config is only accessible during `context` time, thus computing with this may lead to problems.
+> Note, that the information is stored at slide level, meaning the config you set specifically for the slide you write this function in will be accessed, not the global config, outside of the slide.
+
+```example
+>>> #import "@preview/touying:0.6.3": *
```
OrangeX4: Can be changed to 0.7.0

docs/zh/faq.md
```diff
@@ -822,6 +822,38 @@ Content here.
 ]
 ```
 
+### 如何访问全局或幻灯片的配置信息？
+在需要获取配置的位置使用 `touying-get-config`。由于配置只能在 `context` 时机访问，因此基于此进行计算可能会导致问题。
+> 请注意，配置信息存储于幻灯片级别，即访问到的是你为当前幻灯片专门设置的配置，而非幻灯片外部的全局配置。
+```example
+>>> #import "@preview/touying:0.6.3": *
```
OrangeX4: Can be changed to 0.7.0

src/core.typ
zral0kh: well i feel like we could just impose some format onto that.
and actually i am in favor of allowing `touying-get-config()` which enables the user to access stuff like so:
`touying-get-config().info.author` if they want to.

the string being passed in is then just another way to access the specific config directly and since its syntax overlaps with the typst syntax i feel its ok to impose this syntax.

for the non ascii text characters yes that is true, we cannot know how custom themes set their own stuff...
might need to lookup the unicode ranges for text characters of those languages? maybe there is a regex shorthand for that as well?

src/core.typ
zral0kh: yes i also thought of that, typst also throws an error when the key is not present so perhaps we should do the same.

src/core.typ
zral0kh: these do feel somewhat generic so the bindings might overlap with other packages' bindings.

we have many bindings that carry the `touying-` prefix bc we prefer the wildcard import (i agree it is nicer for a package you use a lot like touying), e.g. touying-recall, touying-slide, touying-fn-wrapper

but also we have many that are short. i think since set-config and get-config are not called that often its okay with the prefix

docs/en/faq.md
zral0kh: should we do this for all examples? 
i mean we could even have it be 0.7.1 if this will only go live with the next patch

and we should probably have some tool that sets all these places with versions when we do a new release.

src/core.typ
OrangeX4: I think this idea `#touying-get-config().info.author`  is great, I wrote a demo code:

```typst
#let touying-get-config(..args) = {
  if args.pos().len() == 1 {
    // TODO
  } else if args.pos().len() == 0 {
    let _get-fn-maps(keys, config) = {
      if type(config) == dictionary {
        let maps = (:)
        for k in config.keys() {
          maps.insert(k, _get-fn-maps(keys + (k,), config.at(k)))
        }
        return maps
      } else {
        touying-fn-wrapper((self: none) => {
          let final = self
          for cur in keys {
            final = final.at(cur)
          }
          return final
        })
      }
    }
    return _get-fn-maps((), default-config)   // default-config from configs.typ
  } else {
    panic()
  }
}
```

Indeed, we can write a complex regular expression, but if for simplicity, removing the assert is also feasible.

src/core.typ
zral0kh: true i just did not think of that. but this now just returns a raw inline typc value. not a lot better

e.g. `let d = (a:2); #repr([#d])` prints
`raw(text: "a:2", block: false, lang: "typc")` into the document, this does not solve the dictionary type preservation

src/core.typ
zral0kh: not entirely sure about your demo code, since you call _get-fn-maps just with the default config. We can in reality just return self if no key is inputted.

and the problem is when we push this to slide parts it has to be content, nothing else, otherwise it cannot be summed into a sequence to be pushed into a slide.

to be able to still access the dict afterwards we either need to just write a standard context expression and not have it go through to being parsed and outputted or we reconstruct the dict from the output at a later stage, e.g. be recognizing the output via some label.

src/core.typ
OrangeX4: Dictionary type preservation is impossible, but I just think we can use `#let s = "some text"; #[#s]` instead of `#let s = "some text"; #repr(s)` to avoid `s` was rendered as `"some text"`, while `some text` is expected.

docs/en/faq.md
OrangeX4: We can first use the current version, such as 0.7.0, to make it easier for us to replace all instances of version numbers later. As for the tool, a simple text search and replace will suffice, and this won't be very time-consuming. We also don't need to worry about it appearing prematurely on the documentation website, because we won't immediately push the docs directory to the website.

src/core.typ
OrangeX4: Because we can't directly return self, touying-get-config can only immediately return a dictionary containing many callback functions.

src/core.typ
zral0kh: yes thats true.
well i won't say dictionary type preservation is impossible its just hard

src/core.typ
zral0kh: ahh now i understand your idea. we emit the dictionary with all leaf values replaced by the fn wrapper.

now this still has the problem that it only works for leaf keys. maybe we can introduce some syntax which populates the dict with keys like `info_` or `info-dict`, which holds the whole directory for that key.
Intermediate keys won't be accessed by users often anyway, this should be fine.

and for the possibility when there is one arg, well we just trace through the default config as far as possible, emit a warning when the key is not in there (possibly bc its a custom key inside store introduced by some theme), and then write a callback into the touying-fn-wrapper that will try to parse the remaining thing at context time. 

i think its great, it allows us to make getting touying config feel almost typst-native and custom stuff is also there if you want it. perhaps we can tell custom themes to register their additional fields with touying so that we can also natively support those, could just be one parameter, e.g. `register-config: (custom-key: def-value)` or some callback function they can call like `register-config( addition-config )` and this will be put onto store.

zral0kh: i didn't get it since i thought we emit functions that will have to be called, but because typst calls all functions even without `()` natively this looks good.

zral0kh: i now think we don't need to return any intermediate stuff as those are `repr` representations of the config and typst itself says you should not count on it being an inverse to `eval`. Therefore its good that the result is `content` and we don't want to encourage using it weirdly. If people really need it, they can build it themselves.

and for the config registering, well default-config already has the store which is populated by the theme so it naturally works.

PR #336, opened 2026-04-02
feat: `breakable` and `clip` options to avoid slide overflow

### LINKED ISSUE #335: feat: emit an error/warning when slide content overflows
Currently, when the content in a section does not fit into one slide, they automatically overflow to the next slide. It'd be great to allow an option to prevent this and emit an warning or error when compiling, indicating which slides and code lines are affected.

This is important because in an agentic workflow, it's hard for agent to check the slides visually if one piece of source is mapped to multiple slides and the mapping is messed up (e.g. slide 5 in source code mapped to slide 7 and 8(, this confuses the agent. If a warning is emitted, agent can deal with it on their own.

OrangeX4:
- breakable (bool): Whether to allow slide content to overflow to the next page. When `true` (default), content that exceeds the slide height will automatically create new pages. When `false`, content is constrained to a single page using a non-breakable block, which is useful for ensuring a strict one-to-one mapping between source slides and output pages in agentic workflows. Default is `true`.
- clip (bool): Whether to clip overflowing slide content when `breakable` is `false`. When `true`, content that exceeds the slide height will be visually truncated. When `false`, overflowing content remains visible but does not create new pages. Only takes effect when `breakable` is `false`. Default is `false`.
- detect-overflow (bool): Whether to detect and panic on overflow. When `true`, a `layout` + `measure` check is performed and `panic()` is called if the content height exceeds the available container height. When `false`, no overflow detection is performed (avoids the `layout` overhead). Default is `true`.

Fixes #335


sghng:
If unexpected overflow/clipping happens when breakable set to false, it would be great if a clear compilation warning/error is emitted. I think warnings are more helpful.

OrangeX4:
Perhaps we can consider how to measure overflow; I haven't quite figured it out yet, but maybe `measure` will help.

OrangeX4:
Detected overflow and magic warnings.

image

src/components.typ
```diff
@@ -654,3 +655,71 @@
...
+    radius: (:),
+    spacing: 0pt,
+    sticky: false,
+    stroke: (:),
+  )
+  if detect-overflow {
+    // Detect and warn on overflow
+    layout(container-size => {
+      let content-size = measure(block(
+        ..default-block-args,
```
zral0kh:
```suggestion
        ..tight-block-args,
```

src/components.typ
```diff
@@ -654,3 +655,71 @@
...
+/// - detect-overflow (bool): Whether to detect and panic on overflow. When `true`,
+///   a `layout` + `measure` check is performed and `panic()` is called if the
+///   content height exceeds the available container height. When `false`, no
+///   overflow detection is performed (avoids the `layout` overhead). Default is `false`.
+///
+/// - body (content): The slide content to constrain within a single page.
+///
+/// -> content
+#let page-container(self: none, clip: false, detect-overflow: false, body) = {
+  let default-block-args = (
```
zral0kh: rename to `tight-block-args` since thats what is does.
```suggestion
  let tight-block-args = (
```
then also update below, see the two instances.

src/components.typ
```diff
@@ -654,3 +655,71 @@
...
+            + ", available height: "
+            + repr(available-height)
+            + ").",
+        )
+      }
+    })
+  }
+  // Disable breakability to prevent overflowing content from creating new pages
+  block(
+    ..default-block-args,
```
zral0kh:
```suggestion
    ..tight-block-args,
```

src/magic.typ
```diff
@@ -1,4 +1,16 @@
...
+// Warning
+// ---------------------------------------------------------------------
+
+
+/// Display a warning message with `set text(font: ..)` magic.
+///
+/// - message (str): The warning message.
+///
+/// -> content
+#let magic-warning(message) = {
```
zral0kh:
```suggestion
#let warning(message) = {
```

rename to only `warning`, since we export the magic module as is
see also change above in components

src/components.typ
```diff
@@ -1,4 +1,5 @@
 #import "utils.typ"
+#import "magic.typ": magic-warning
```
zral0kh:
```suggestion
#import "magic.typ": warning
```

src/components.typ
```diff
@@ -654,3 +655,71 @@
...
+    layout(container-size => {
+      let content-size = measure(block(
+        ..default-block-args,
+        width: container-size.width,
+        body,
+      ))
+      let content-height = content-size.height
+      let available-height = container-size.height
+      if content-height > available-height {
+        magic-warning(
```
zral0kh:
```suggestion
        warning(
```

zral0kh (changes requested):
yes nice one, good feature regardless of agents.
and i really like the warning.

somehow it does not show all the little things i'd like to rename/change, please have look

zral0kh (approved):

PR #339, opened 2026-04-03
feat(components): add `lazy-v` (`lazy-h`) and `lazy-layout` for equalizing multi-column (-row) block heights (widths)

OrangeX4:
## Summary

Introduces two new components, `lazy-v` and `lazy-layout`, that together
solve a common layout problem in multi-column slides: making all columns
the same height while keeping the overall container height equal to the
tallest column (not the full page height).

## Problem

When placing multiple blocks side by side (e.g. via `side-by-side` or a
manual `grid`), there is no straightforward way to push footer content
(such as a caption or label) to the bottom of each block and have it
align across all columns. Using `v(1fr)` directly causes the block to
expand to fill the entire page height, which is not the desired behavior.

## Solution

- `lazy-v(amount)`: A deferred fractional spacer. Place it between the
  main content and the footer inside a block. It is invisible during
  height measurement and only activates when rendered inside `lazy-layout`.

- `lazy-layout(body)`: Wraps a multi-column layout. It first measures
  the natural height of the content (ignoring all `lazy-v` markers), then
  re-renders at that fixed height with the markers activated. This causes
  each block to stretch to match the tallest sibling, without the container
  growing beyond the tallest block's height.

## Usage

```typ
// Via side-by-side (recommended)
#components.side-by-side(lazy-layout: true)[
  #block(width: 100%)[
    #lorem(10)
    #components.lazy-v(1fr)
    Bottom left.
  ]
][
  #block(width: 100%)[
    #lorem(20)
    #components.lazy-v(1fr)
    Bottom right.  // aligns with "Bottom left." above
  ]
]


zral0kh:
could you also introduce lazy-h, mainly for parity.

i know we can theoretically always use grids but the same holds for achieving lazy-v layouting.

then `lazy-layout` would get a `direction` argument, which allows a typst `direction`: one of `ttb, btt, ltr, rtl` and inside we simply use the `direction.axis()` to distinguish.
alternatively we accept `"horizontal"`, `"vertical"` directly, 

or `lazy-layout` does this distinction automatically based on the first lazy-space function it encounters, or we panic when it encounters different lazy space functions.

OrangeX4:
done.

zral0kh:
i noticed that lazy-h slightly pushes content outside the box:
grafik

when i set inset to 0pt. this is probably a bug from typst.
is does occur with other insets as well, it is just not as easily observable there.

it seems typst is no longer correctly counting spaces.
i don't see a way to fix that, the behaviour also appears if you just use `h(1fr)`.

thus i guess its done.

zral0kh (approved):

PR #340, opened 2026-04-03
improved warning via unicode backspace

zral0kh:
uses unicode backspace to delete the old font warning, and notes that this is a warning from touying.

quick one, but i randomly had the idea and this is much nicer.

### DIFF
--- a/src/magic.typ
+++ b/src/magic.typ
@@ -8,9 +8,9 @@
 /// - message (str): The warning message.
 ///
 /// -> content
-#let warning(prefix: "[touying]", message) = {
+#let warning(prefix: "[touying] ", message) = {
   let delete-old-warning = range(21).map(i => "\u{0008}").sum()
-  set text(font: delete-old-warning + prefix + " " + message)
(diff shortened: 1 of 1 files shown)


zral0kh:
btw hopefully the core typst maintainers don't see this, they don't want packages to use warnings rn as typst has no way to block them and they fear warning spamming.

src/magic.typ
```diff
@@ -9,7 +9,8 @@
 ///
 /// -> content
 #let warning(message) = {
-  set text(font: "[WARNING] " + message)
+  let delete-old-warning = range(21).map(i => "\u{0008}").sum()
+  set text(font: delete-old-warning + "[touying] " + message)
```
OrangeX4: Considering that some users might want to use this, perhaps we don't need to mark `[touying]`?

zral0kh: maybe we can set this as some field, e.g. "prefix" or so.
and by default it is `[touying]`.

just don't want to write it out in all locations where we might call this. and also i think it's helpful to know touying's warnings are not typst core warnings.

Issue #224, opened 2025-10-16
Add email to the config-info and title slide


Clark-Young:
During academic presentations, speakers may want to include their contact information, such as their email address, on the cover. I think it's necessary to add this funtion for touying.

zral0kh:
in favor of this, maybe we can also introduce some field called `extra` or similar which enables users to set arbitrary extra info at the bottom, or wherever the theme places it.

Could be used e.g. for a url, telephone number, ..., 

maybe extra can also be a dict and normal themes just display all the values, but specialized themes may use the extra fields like normal info fields and extract values individually and place them whereever they want.

Edit: yes theme specific stuff should normally be constrained to `config-store`, but i feel like this is one of the exceptions to that rule.

Example: Some theme may want to display authors with different institutions and contacts. This could be done via the `extra` field. You can set `extra: (author1:(name: "Author 1", institution: "First Inst.", contact: "author@first.edu"))` and some special theme could utilize this.

PR #342, opened 2026-04-04
Feature: Additional `contact` and `extra` field in `config-info`

zral0kh:
As hinted by #224 additional info is often required.
This introduces a general `contact` field and some `extra` for stuff certain themes/users might need.

### DIFF
--- a/src/configs.typ
+++ b/src/configs.typ
@@ -531,6 +531,7 @@
   institution: _default,
   contact: _default,
   logo: _default,
+  extra: _default,
   ..args,
 ) = {
   assert(args.pos().len() == 0, message: "Unexpected positional arguments.")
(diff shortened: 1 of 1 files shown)


src/configs.typ
```diff
@@ -524,6 +529,7 @@
   author: _default,
   date: _default,
   institution: _default,
+  contact: _default,
   logo: _default,
   ..args,
```
OrangeX4, line 535: Is extra missing here as an input parameter?

zral0kh:
@OrangeX4 have a quick look again please, i updated some docs.

src/configs.typ
zral0kh: yeah idk how i forgot that, should be there now

OrangeX4 (approved):

Issue #343, opened 2026-04-05
Feature: Add ToC slide before each section, with current section highlighted


yjcui:
Posted this feature request a year ago under 'Discussions', which receives no discussion or response. Repost it here and hope someone can take a look and comment. This is a very useful feature from LaTeX Beamer.

https://github.com/touying-typ/touying/discussions/176

OrangeX4:
You are currently able to use a similar ability on the dewdrop theme, further, I think this issue should be a duplicate of #137, maybe you can refer to it

yjcui:
Thanks for the response! It seems a duplicate of #137. Still it makes more sense if we can have this useful and common feature 'inherent' of Touying package itself rather than being limited to certain themes, or relying on 'hacky tricks' to create, if this can be implemented with reasonable efforts.

Again great thanks for this package!

Issue #137, opened 2024-12-12
Outline slides for each subsection


ThijsJenneskens:
Hi! I use Touying a lot to make presentations :) thank you for developing it, it has done a lot for me.

I would like a feature where I can automatically have a mini-outline for each section, showing the subsections that will come up in this section. In that way, I can give my audience a very short preview at the beginning of every section. Is there a way to automatically do this?

Thanks,
Thijs

eusebe:
Related to [this](https://github.com/touying-typ/touying/issues/56#issuecomment-2343638597) ?

enklht:
While it's not fully automated yet, I managed to create an example that might help. You can check it out here: https://typst.app/project/wW4El9k64GTdQPe5WZRDyu.

I hope this helps.

Thijsie2:
@enklht Awesome! thanks for the quick reply :)

I think @eusebe's [request](https://github.com/touying-typ/touying/issues/56#issuecomment-2343638597) is still valuable to take into account for a future version of Touying.

enklht:
I think improving the custom outline is a good idea.
I will try to implement features like Beamer's `sectionstyle` and `subsectionstyle`. Thanks!

PR #345, opened 2026-04-08
Improved `custom-progressive-outline` and new `section-relationship` and some other things

zral0kh:
#137 and #343 wanted to be able to have varying depth outlines depending on where in the document we are.
This is now achievable in the custom-progressive-outline via `show-past`, `show-current`, `show-future`, which are boolean arrays (one for each level) that toggle this behaviour. Further if a function is passed this acts as a style function for the entry which allows having different cover styles for past and future sections.

The third page of the new test "features/custom-outline"

It is created with this `custom-progressive-outline` call:

```typst
#components.custom-progressive-outline(
      self: self,
      level: 1,
      show-past: (true, false),
      show-future: (true, false),
      show-current: (true, true, false),
      vspace: (.0em, .0em),
      numbering: ("1.1",),
      numbered: (true,),
      title: none,
)
```

I deprecated the `uncover-fn` as it was not consistent with the rest and instead we now support text-style (array of dicts: for all text styles) and current-style (which overrides that when heading is not covered). `uncover-fn` is now deprecated but `text-fill`, `text-size`, `text-weight` still work.

Bc we want to keep some similarity to old behaviour this might not be enough for some occassions since the cover mechanism depends on the weak before/after section range calculation with a level, thus we now also support `utils.section-relationship`, which returns a more detailed relationship as an int (see the docstring for more detail).
We might want to consider instead using an array there, as that would give full customizability, but i could not think of scenarios where it would be necessary to know the exact distance in terms of section count per level.

with that we can achieve stuff like this: (test, page 5)

with the code looking like this

```typst
#show outline.entry: it => {
    let relationship = utils.section-relationship(it)
    let current = utils.current-heading()
    let alpha = if relationship == -2 or relationship > 0 { 40% } else { 100% }
    let weight = if relationship == 0 and current.level == it.level {
      "bold"
    } else { "regular" }
    if it.level > 1 and calc.abs(relationship) > 1 {
      text(fill: red, it)
    } else {
      text(fill: utils.update-alpha(text.fill, alpha), weight: weight, it)
    }
}
#outline(title: none)
```

And we might want to write in the docstring of  `custom-progressive-outline` which parameters have an autorepeat for their last entry, or which ones don't: `vspace` and `numbering`

And the minor things: updating the themes to the new api, improving warning, updating the warning and docstring of the page thing where we detect overflow.


OrangeX4 (approved):

OrangeX4:
Thanks!

PR #346, opened 2026-04-08
New theme: Cinema

philocalyst:
- New theme, Cinema!!!
- Employs a large number of visual effects
- Added a photo to demonstrate noise functionality.


zral0kh:
First of, thank you very much for contributing to touying.

Next, why do you think this theme should be added to touying core and not live outside as a separate package like the many other touying themes in typst universe?

And a minor thing: relying on pngs for the noise may cause some pdf viewers to not correctly scale the image and you get pixel artifacts. This is not bad when the noise is soft anyway, but sharp noise may yield to unintended looks.

philocalyst:
PNG is good advice.

I really wanted to contribute directly because I think the theme is representative of the absolute flashiest Touying slideshow, and that's meaningful and important. When looking through the examples I didn't see anything that focused on "beauty" (Whatever that means).

I went ahead and tried to do that!

I think it would be a fun theme to experiment with the bleeding edge -- like if Typst makes it easier to create noise, for example.

Or if we wanted to add multiple slide formats.

▰▰▰▰▰
Miles Wirth 🙃

OrangeX4:
I think we should also think carefully about adding a new theme to the touying built-in theme, as it may increase the burden on the main maintainers of touying when updating the version. I think this theme looks good, thanks for your contribution. But it seems to use a lot of gradients or some other technology, and the overall rendering will be slower, or there are certain performance issues? So considering the above two points, maybe a separate template would be more appropriate, which would make the touying's built-in themes more concise. Of course, this is just my opinion at the moment.

philocalyst:
It's heavier, for sure -- but only by a smidge, typst handles it really well!

I think if the goal with Touyings builtin themes is to showcase then this my point stands, the more diversity in presentations available the more battle-tested major refactors will be imo.

▰▰▰▰▰
Miles Wirth 🙃

zral0kh:
I get your point, and i don't even really care about how heavy the theme on the compile time is, but adding more themes just serves as extra burden to the core maintainers, whose aim is functionality, not pretty themes. 

Whenever we make changes to themes and their api, we publish migration guides and the themes in touying core serve as an example, but that means we need to keep all those themes inside touying core up to date. Yes it also is testing ground, but the more themes there are the lesser the coverage improvement becomes, especially if they rely on basic touying functionality, that is already covered to a very large part. This theme does not seem to use touying functionality in any non-basic way.

In a perfect world we could migrate all themes out there and have sth like a touying theme store which is maintained and kept roughly up to date with touying, but that is just too much work. We'd much rather have the community keep their themes up to date and ask here when they have questions.
While that means many people using touying themes from universe will get features later, there is nothing really we can do about it. Currently we have 36 themes popping up in universe when you search for touying under templates. There are probably a couple more that are not found by that keyword.

philocalyst:
Would a neat middle-ground be a Touying org? Where individual maintainers could have admin or write perms on their own themes, and encourage direct communication with the main Touying project?

That way things could be maintained on a best-effort basis, if people fall off then it would be easy to assign new maintainers, and people could expect a decent level of staibility. The official themes could also be moved and then be the up-to-date "official examples"

▰▰▰▰▰
Miles Wirth 🙃

OrangeX4:
Personally, I'm not really against adding new built-in themes to touying, but the premise is that this theme should be relatively simple, not overly stylized, and not too difficult to maintain.

Based on this principle, I believe it's better to have a separate template package for cinema themes, because:
1. The cinema theme relies quite heavily on a png image to add noise for optimal effect, and this png image isn't easy to place directly into the touying package source code (in the src directory), making it difficult to use copy-and-paste code.
2. The background color for the cinema theme is black. To achieve the best effect, it requires separate adaptation for many third-party packages, such as fletcher and theorion; otherwise, the actual effect will not be too good, which requires a relatively large amount of theme-specific maintenance.
3. Cinema is relatively stylistic; if you need to further add stylistic adaptations (such as selecting a special default art font), it's more convenient to maintain it as an independent template.

image


Based on the above reasons, I would prefer cinema to be released as a template package, making it easier to discover on typst universe. If you like, you can also add your demo to the touying GitHub wiki page to help more people discover it.

philocalyst:
You never got back to me about the organization idea?

▰▰▰▰▰
Miles Wirth 🙃

OrangeX4:
I don't think this PR will be merged after you join the organization, but you will join the organization after providing some PRs that can be merged to touying, which is a priority issue. What I'm considering here is that this theme may not be suitable for merge as a built-in theme for reasons like I mentioned above, so it's not that you joined the organization, this PR is will suitable to be merged. Of course, if you propose a few PRs that are suitable to be merged to touying and successfully merge, you are certainly welcome to join.

philocalyst:
You're misunderstanding. Please read back on my org proposal to host some of these external themes.

▰▰▰▰▰
Miles Wirth 🙃

OrangeX4:
You mean create a new repository under the touying-typ organization, such as `touying-typ/cinema-theme`? At the moment I don't feel like it's much different from templates from not touying-typ organizations, and I don't see any particular benefits here...

philocalyst:
I don't want to keep pressing this, but I did outline benefits above, but maybe you are right they mean little.

▰▰▰▰▰
Miles Wirth 🙃

zral0kh:
i get what you want:

to have your theme be part of something official.

but rn we don't feel like we have enough capacity to add large new themes or even start such a big new part of touying, bc then people expect that to be maintained. And that for years to come...

sorry

philocalyst:
Got it, just wanted to feel like you understood. Thanks for explaining :)

▰▰▰▰▰
Miles Wirth 🙃

PR #348, opened 2026-04-08
improved mark-warning, and now also displayed if false as a magic.warning

zral0kh:
improved the warning message of the mark-warning and now also emit it via magic.warning when we don't panic, to not silently exit.

### DIFF
--- a/src/configs.typ
+++ b/src/configs.typ
@@ -61,13 +61,18 @@
     let marks = query(<touying-temporary-mark>)
     if marks.len() > 0 {
       let page-num = marks.at(0).location().page()
-      let slide-name = query(selector(heading).before(marks.at(0).location())).last().body.text
+      let slide-name = query(selector(heading).before(marks.at(0).location()))
+        .last()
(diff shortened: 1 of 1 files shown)


zral0kh:
perhaps when we continue to use the magic.warning we should give users the ability to turn them off.

zral0kh:
perhaps we could implement a `magic.disable-warnings` as the following

```typst
//magic.typ
#let disable-warnings(body) = {
   let random-number = BIG_NUMBER
   set text(features: (tyng: random-number, nwrn: random-number))
   body
}
//user-doc
#show: magic.disable-warnings
```

update: warnings package will soon be published as `uniwarn`:  https://github.com/typst/packages/pull/4542

i guess they won't have a problem with it, they probably have discussed it internally by now.

zral0kh:
wait until #359

zral0kh:
done @OrangeX4 review this little change pls

OrangeX4 (approved):

PR #349, opened 2026-04-09
Fixes #81, metropolis outline not  being the same as the beamer theme.

### LINKED ISSUE #81: Metropolis theme - Outline not consistent with beamer theme
The pictures and the code given for the `metropolis` theme don't correspond since the code:
```typ
= Outline <touying:hidden>

#outline(title: none, indent: 1em, depth: 1)
```
gives the correct outline but with slides numbers (see picture below), which is intended but not fit the beamer theme.
image

A very simple workaround would be to include in the `init` function, the following piece of code:
```typ
#show outline.entry: it => {
  block(above: 0pt)[#it.body]
}
```

Adding these lines produces
image

Cheers !

zral0kh:
This introduces an outline slide for metropolis, which fixes #81.


OrangeX4:
lgtm, but should the examples and docs be updated?

OrangeX4 (approved):

PR #351, opened 2026-04-13
docs: larify FAQ heading about auto TOC page

ph3n92h3:
Update the FAQ headings in English and Chinese to more precisely describe the issue: replace the vague "disable automatic section slides" heading with wording that explains preventing a highlighted table-of-contents page from appearing before the top-level title. The example/config remains unchanged (config-common(new-section-slide-fn: none)). Files updated: docs/en/faq.md, docs/zh/faq.md.

### DIFF
--- a/docs/en/faq.md
+++ b/docs/en/faq.md
@@ -556,7 +556,7 @@ Use `#set par(justify: true)`:
 
 ## Headings and Sections
 
-### How do I disable automatic section slides?
+### How do I prevent an highlighted table of contents page from automatically appearing before the top-level title?
 
 Set `config-common(new-section-slide-fn: none)`:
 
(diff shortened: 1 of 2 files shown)


zral0kh:
this does seem better than before but i still think how people search for specific features really is up to how they expect stuff to be named, which may be totally different from your expectation.

That's why doc helper llms are so useful. But i understand most people don't use those (most of the time i also don't use those). Maybe we can instead start providing keywords/-terms for docs and faq.

OrangeX4:
Personally, I think this pr is less clear, and `automatic section slides` better reflect what it is.

OrangeX4:
For the reasons mentioned above, I'm temporarily closing this PR. We can reopen it if there are any further comments.

PR #352, opened 2026-04-13
Feature: #154 Left Right Navigation

### LINKED ISSUE #154: Navigation buttons
I miss those little Beamer navigation buttons when I'm using Touying.

Image

I tried to reproduce them with:

```typ
#context {
  let cur = query(selector(heading).before(here())).len()
  let prev = query(selector(heading)).at(cur - 2)
  let next = query(selector(heading)).at(cur)
  [
    #link(prev.location())[⬅️] #link(next.location())[➡️]
  ]
}
```

This doesn't work because this method doesn't skip animation steps.

zral0kh:
This introduces the feature wanted by #154 and thus completes it.

It further provides new docs on navigation elements and updates the deprecated counter.display syntax from the docs.

We might want to think about how we name the thing exactly and also the parameters of the function, feel free to change stuff.


zral0kh:
@OrangeX4 btw could you update the zread?

src/core.typ
```diff
@@ -1147,6 +1154,110 @@
...
 
+
+/// Create a navigation element with links to previous/next subslide/slide and a center icon. Note that physical pages are considered subslides, thus your content breaking the slide into multiple pages will be considered as 'multiple' subslides.
+///
+/// - icon (symbol, content): The icon to display for the navigation.
+/// - nav (symbol, dictionary, content): The navigation symbols. We use filled symbols as the links to the subslides and stroked symbols as the links to the pages. You can also pass a dictionary with the structure `(filled:(left:any, right:any), stroked:(left:any, right:any))` and you may omit arbitrary fields. We try the best we can to fill the missing fields based on the values and symbols provided.
+/// - mode (str): The mode of the navigation item. Can be "both", "subslide" or "page".
+/// - show-useless (bool): Whether to show the navigation links when they are useless (e.g. on the first page, the "previous page" link is useless). Default is `true`.
+/// -> content
+#let lr-navigation(
```
OrangeX4, line 1176: Why not put it in the components file?

OrangeX4: Perhaps we could add some test files.

docs/en/tutorials/progress/counters.md
```diff
@@ -11,8 +11,8 @@ Touying provides a set of counters and utilities for tracking and displaying pre
 `utils.slide-counter` is the primary Typst counter that increments on every slide.
 
 ```typst
-// Display the current slide number
-#context utils.slide-counter.display()
```
OrangeX4, line 15: Is it necessary to make changes here?

docs/zh/tutorials/progress/navigation.md
```diff
@@ -0,0 +1,91 @@
...
+# 导航元素
+在较为复杂的幻灯片主题中，加入导航元素是很常见的做法。例如可以参考 [dewdrop 主题](https://touying-typ.github.io/docs/themes/dewdrop)。
+## 左右导航
+[`lr-navigation`](https://touying-typ.github.io/docs/reference/core/lr-navigation) 可创建可点击的上一页与下一页控件，支持按幻灯片、子幻灯片（物理页面）或两者同时导航。
+```example
+#import "@preview/touying:0.7.1": *
+#import themes.simple: *
+#show: simple-theme.with(
+  footer: self=>[
+    #align(center,
```
OrangeX4, line 18: It's best to format it before adding the document.

OrangeX4:
btw I've already triggered a zread update.

docs/en/tutorials/progress/counters.md
zral0kh: this was outdated, we can no longer use `display`, but need to use `get` for counters.

src/core.typ
zral0kh: sadly putting in components causes circular dependencies.
we need the `touying-fn-wrapper-raw` to allow it without passing in self, and that one is defined in core

could possibly fix that in the refactor ...  "maybe"

docs/zh/tutorials/progress/navigation.md
zral0kh: done

src/core.typ
zral0kh: since we now expose a couple components directly i think its fine... (not pretty tho)

zral0kh: we could do an lr-component that requires self always and put it into components, just like we do with `uncover` both in utils and core...

OrangeX4 (approved):

PR #354, opened 2026-04-15
feat!: always attach `#speaker-note[]` to the previous slide & default `receive-body-for-new-*-slide-fn` to `false`

### LINKED ISSUE #240: [Dewdrop theme] Content in first level of header is ignored
If we have content such as simple text under a first level of heading but not a second level of heading, this content is simply ignored and not part of the PDF (slides aren't created). However the first level of heading appears in the navigation mini-slides and the outline, but obviously leads to nowhere.

Issue encountered with the Dewddrop theme. Stargazer isn't affected and I didn't tried other themes.

OrangeX4:
**⚠️ Breaking Change**

`#speaker-note[]` now always attaches to the **slide above it**, regardless of how that slide was created:

- Explicit slide calls: `#slide[...]`, `#title-slide()`, `#outline-slide()`, etc.
- Heading-triggered section slides: `= Section Title`
- Normal content slides: `== Subsection` with body content

This eliminates the common pitfall where a `#speaker-note[]` placed after a slide would silently create an unwanted empty "ghost" slide.

As a consequence, `receive-body-for-new-section-slide-fn` and its variants are now **defaulted to `false`**. Previously these were `true`, which caused content after `= Section` headings (including `#speaker-note[]`) to be absorbed into the section slide's body — a behavior that was often confusing and rarely intended. With speaker-notes now correctly attaching to the previous slide, this workaround is no longer needed.

**What changed**

1. **`src/core.typ` — `split-content-into-slides`**:
   - Tracks the last emitted slide (both explicit `#slide[]` wrappers and heading-triggered section slides) via `last-wrapper-info`
   - When a `#speaker-note[]` immediately follows a slide with no other content in between, it injects the note into `self.attached-speaker-notes` and re-generates the slide — instead of pushing it into `slide-parts` where it would trigger a ghost flush
   - Automatically clears the tracking when non-whitespace, non-speaker-note content is encountered

2. **`src/core.typ` — `touying-slide` → `subslide-preamble`**:
   - Processes `self.attached-speaker-notes` inside the slide's page context, ensuring both pdfpc metadata and `slide-note-state` updates work correctly with `show-notes-on-second-screen`

3. **`src/configs.typ`** — Default values changed:
   - `receive-body-for-new-section-slide-fn`: `true` → **`false`**
   - `receive-body-for-new-subsection-slide-fn`: `true` → **`false`**
   - `receive-body-for-new-subsubsection-slide-fn`: `true` → **`false`**
   - `receive-body-for-new-subsubsubsection-slide-fn`: `true` → **`false`**

**Migration guide**

If you relied on content after `= Section` headings being absorbed into the section slide body, explicitly set `receive-body-for-new-section-slide-fn: true` in your `config-common(...)`.

**Tested scenarios**

| Scenario | Result |
|---|---|
| `#slide[...] + #speaker-note[...]` | ✅ Note attaches to previous slide |
| `#slide[... #pause ...] + #speaker-note[...]` | ✅ Note appears on all subslides |
| `#slide[...] + multiple `#speaker-note[...]` | ✅ All notes attach |
| `= Section` + `#speaker-note[...]` (metropolis, receive-body: false) | ✅ Note attaches to section slide |
| Inline `#speaker-note[...]` (existing behavior) | ✅ Unchanged |
| `show-notes-on-second-screen` panel | ✅ Notes render correctly |
| All existing tests & theme examples | ✅ No regressions |


zral0kh:
nice solution

zral0kh:
should this get onto an v0.8.0 branch? since it's breaking? 
will look thoroughly through it later.

i'd like to postpone v0.8 until i have the refactor and docmode ready, not sure when i have that time, couple of weeks maybe

OrangeX4:
I've recently been trying to use AI to assist in writing touying slides, so I hope to iterate quickly and implement more AI-friendly features. I personally hope to merge the current open pull requests tonight and release a new version tomorrow (either 0.7.2 or 0.8.0 is fine; I don't think it's a huge breaking change).

I also have another idea: to export some components to the top level, so we can directly use syntax like `#side-by-side[][]` instead of having complicated prefixes. With these all changes, we can achieve a syntax that is very AI-friendly, avoid many potential problems, and provide a concise api.

zral0kh:
i'm okay with this being in touying 0.7.2, and not totally opposed to the idea of more openly exposing the components, but exposing more and more stuff will eventually cause clashes with other packages so i'm not the biggest fan.

if you feel like writing slides with AI is a big thing you want, then maybe we can just provide some touying.md skill file for llms to use. this makes it both ai and human friendly.

if you'd want you could do:

```typst
#import "@preview/touying:0.7.1":*
#import components: *
#import themes.simple:*
...
```

nothing hinders you from doing that, that is one line for the entire document.

i do agree that we could get rid of some of the `touying-` prefixes, those seem a bit weird, but the problem stays: since we encourage using the wildcard import many people do that and thus our naming should be explicit. e.g. for the config i think having it be `touying-set-config` and `touying-get-config` is alright (but could also be  `slide-Xet-config`, maybe also `Xet-slide-config`) . same for overridden definitions like `touying-mitex`, etc.

the `touying-slide-wrapper` and `touying-slide(s)` should also stay for compatibility reasons and just bc i think it makes sense since only theme creators need these basically, not the average user.

the `touying-fn-wrapper` is the same as `touying-set-config` i believe. BTW we don't export the `touying-fn-wrapper-raw` rn oopsie xD
(if we'd rename to the config stuff to `Xet-slide-config` these could become `with-slide-config(raw:bool, self=>(...))`)
and ofc old names would stay for quite a while.

just some thoughts on the topic...

src/core.typ
```diff
@@ -288,6 +288,11 @@
   // leading preamble to collect content from before the slide break
   let leading-preamble = ()
 
+  // Buffer for the last slide-wrapper's callable and self, so that
+  // immediately following speaker-notes can be re-attached to it.
+  // Each element is a (callable, slide-self) pair. Empty means no pending wrapper.
+  let last-wrapper-info = ()
```
zral0kh, line 294:
```suggestion
  let last-wrapper-annexe = ()
```

this makes it more explicit what it is for and easier for future devs to reuse for additional features.

maybe we can just name it `annexe` or `attachments`

zral0kh: nvm i misunderstood the purpose this is fine.

zral0kh: we save the wrapper self so that we can make attachments when necessary, nothing more. its good.

OrangeX4:
I think these `touying-xxx` can be kept as they are, and there's no problem with that. I believe keeping most components in components is correct. I now just think **a few** need to be highlighted separately: `side-by-side` (I hope to add an alias `cols` for brevity and ease of use), `lazy-v`, `lazy-h`, `lazy-layout`. These implementations can be directly called with `#lazy-v(1fr)` for top-level invocation, and the rest can remain as is. Of course, the usage like `#components.side-by-side` will also be retained.

PR #362, opened 2026-04-22
export the raw touying-fn-wrapper as well

zral0kh:
exports the `touying-fn-wrapper-raw` via exports.typ

thought this should finally be done, so people can use it as well

### DIFF
--- a/src/exports.typ
+++ b/src/exports.typ
@@ -25,6 +25,7 @@
   speaker-note,
   touying-equation,
   touying-fn-wrapper,
+  touying-fn-wrapper-raw,
   touying-mitex,
   touying-raw,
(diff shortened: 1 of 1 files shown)


zral0kh:
btw: the releases and version here is not up to date

what was the v0.7.3 bug that needed fixing?

OrangeX4 (approved):

OrangeX4:
I hope to update touying README after the merge in 0.7.3. The error occurred when `config-common(breakable: false)` was used together with `lazy-layout` because the `measure` function used internally by `page-contrainer` couldn't handle the complex context and query mechanisms of `lazy-layout`. Therefore, I made a fallback for this.

PR #363, opened 2026-04-22
Feature: (#302) adds `touying-reduce` which automatically looks up the bindings given a package

### LINKED ISSUE #302: Feature: Automatic Reduce of Cetz and Similar
We can theoretically expose a `reduce` function, which we then could call like

```typst
#reduce(cetz)({
  import cetz.draw: *
  ...
})
```

The only problem is that we need a way to recognize the module so that we can set the reduce and cover function we should normally use.

The best case would be if all such modules would expose a name directly like `cetz.name` which evaluates to "cetz", otherwise we could build patterns on the module layout to and functions defined in it to recognize the correct module.

Then `reduce` would be defined like:

```typst
#let reduce(module) = {
  name = recognize-module(module)
  if name=="cetz" {
    touying-reducer.with(reduce: {module.canvas}, cover: {module.draw.hide.with(bounds: true)})
  } else if ...

}
```

This would make using such modules in touying much more convenient.

We can open issues in relevant modules to expose their name, until then recognize them by module matching and allow package owners to PR their own module to be reduced automatically.

Might have some time in the future for this.

zral0kh:
uses `extern.auto-reducer-bindings` to define the bindings if a package with a field `name` is provided.

also looks into the field `touying-reducer-bindings` of the given package.

currently auto supported packages:
- cetz
- alchemist

if nothing works you can still provide a `bindings` dict. (but in this case it would be better to just predefine the `touying-reducer`.

Other:
- it has a synonym `touying-diagram`


zral0kh:
still requires documentation

zral0kh:
probably would prefer if this becomes the new default, it is simpler for people.

thus a lot of documentation changes.
and sadly fletcher did not do its mini pr to expose its name...

src/core.typ
```diff
@@ -2626,6 +2627,101 @@
...
+      + repr(bindings),
+  )
+  if bindings.reduce == none or bindings.cover == none {
+    //two cases to look bindings up:
+    // 1. in the package itself
+    // 2. in the auto-reducer-bindings based on the package name
+    bindings = pckg.at("touying-reducer-bindings", default: none)
+    if bindings == none {
+      assert(
+        "name" in pckg.keys(),
```
OrangeX4: Maybe we can add a `name: auto` parameter to `touying-reduce`, which also supports `touying-reduce(fletcher, name: "fletcher", {..})`?

zral0kh: that would just be an exception for fletcher

i already added the option for other packages to yield the bindings from within themselves, so i hope we don't have to touch that a lot, but either people make a pr for bindings to be added to the auto bindings or they just add it to their package via `touying-reducer-bindings`.

maybe i can have an llm write a recognition based on the fletcher package layout since they don't seem to maintain it anyways. and if they decide to come back and change stuff they can just expose their name and it also works.

zral0kh: turns out `repr` returns `"<module fletcher>"` for a package named 'fletcher'.

i'd still want to rely on an exposed name as it makes it more clear and officially repr is not meant to be used for such stuff

OrangeX4 (approved):

OrangeX4:
lgtm

PR #364, opened 2026-04-27
Feat easy custom theme and improved `touying-fn-wrapper-raw`

zral0kh:
This PR adds two things:

#### 1. A new `custom-theme`:
- allows for easier theme creation
- abstracts away touyings theme complexity at some customizability loss

#### 2. `touying-fn-wrapper-raw` can now be nested
- its content is now parsed via `_parse-content-into-results-and-repetitions`
- a simple parsing is made for nesting it inside a normal `touying-fn-wrapper`

and fixes one small thing:
- fixed a bug where `pause` and similar did not produce the warning message when inside an animation function, bc their dict did not have an associated function, just the kind. (Also updated the message slightly to be more informative)


zral0kh:
still needs docs ofc

OrangeX4:
Why do we need a custom theme? It doesn't seem to simplify much either?

I'm a bit worried that adding a new custom theme solution might confuse users. Because there are many ways to do the same thing, it's easy to cause confusion for users. Should we perhaps stick with the original custom theme solution and just do it better, rather than adding a new one?

zral0kh:
> Why do we need a custom theme? It doesn't seem to simplify much either?

The thing i wanted to simplify mainly was that people needed to know some touying internals: `touying-slide-wrapper, touying-slide, ...` just to write their own minimal theme.
This is not meant to replace true custom themes.
More of an easy way to quickly throw sth together.

> I'm a bit worried that adding a new custom theme solution might confuse users. Because there are many ways to do the same thing, it's easy to cause confusion for users. Should we perhaps stick with the original custom theme solution and just do it better, rather than adding a new one?

Yes true it might confuse users. The naming is not final ofc. I just thought we might want to have some middle ground between not having your own theme and writing a complex theme yourself. This way it might feel less overwhelming especially when you are just starting writing slides in touying and want to create your own theme.

I still believe we want to keep the current theme customization ability since it allows total control basically, but having something more high level may also make sense.

zral0kh:
Btw i am also not quite finished i have a couple of more ideas floating around:

- using the <label> syntax to mark header and footer containers, which will be recognized as well and put into the page header and footer.
- perhaps reexporting the usual theme building necessary components and functions under a separate `themes.components` module ?

zral0kh:
maybe we can call this `easy-theme` or sth, anyways what do you think currently about this.

My stance is the following: We have theoretically full customizability and people can hook into the same bindings that real themes use, these can however be a bit cumbersome and give too much option for what the average user needs. Yes maybe simplifying this would help but there will always be more complexity than sth we intend to only show the end-user i assume. Secondly changing an existing theme is ofc possible and people can also utilize the setting function to customize the appearance without rewriting a theme, but that does feel a bit weird since you need to overwrite the slide functions with themselves with a setting attached so that it becomes default and then you cannot natively use `slide(setting)` again but must reference it via the theme binding. 

on the other hand perhaps we can wait with something easy until we eventually have the custom types which would allow us to make this much nicer together with other customizability

OrangeX4:
I think easy-theme is a good choice. I agree with other opinions as well, and we can think it through carefully.

PR #370, opened 2026-05-11
Fix #112 by switching to an improved alpha/color changing cover, semi-transparent-cover as fallback

### LINKED ISSUE #112: Semi-transparent-cover doesn't cover everything and renders wrongly

zral0kh:
New approach to fixing #112 .

Alpha changing cover now correctly covers all content types that have accessible colors, and fallback is semi-transparent-cover, it applies an alpha update to all colors.
Styles are read out via context and we apply an inner style node to not throw away possible other styles.

Should work well with normal colors and strokes. (usually fill and stroke)
Gradients/tilings should also work. Gradients are natively supported by the changing covers and tilings get the fallback since we cannot access the content at all.

Color changing cover does cover in a similar way, although less compile time intense since we don't need to call context for it.

remaining tasks:

- [x] implement alpha/color changes for spot colors

remaining fixes:
- spacing between different types of lists during covering

currently used test file:
```typst
#import "/lib.typ": *
#import themes.simple: *
#show: simple-theme.with(
  // config-methods(cover: utils.alpha-changing-cover,),
  config-methods(cover: utils.color-changing-cover.with(color:luma(90%)),),
)

= Title

== Title of slide
$
  f(x) &= pause vec(2x, 2x)  \
       &= pause 2x vec(1, 1)  \
$

== Enums

- double list item #pause
- more double list items #pause
- third marker #pause

1. sth #pause
2. sth else #pause
3. third numbered item #pause

\A Something #pause \
\B Something else #pause \
\C Third item


== Next Slide

Normal Content. #pause Even more normal content.

#figure(
  rect(width: 5em, height: 5em, fill: red, stroke:blue+3pt)[Content],
  caption: "A red square"
)


#text(fill:red)[$f(x) = pause vec(2x, 2x) = pause 2x vec(1, 1)$]

== debug slide

#repr(figure(rect(fill:red, stroke:blue+3pt)[x], caption:"test").fields().keys())

Original rect:
#rect(fill: red, stroke: blue+3pt)[test]

Covered (color-changing-cover, should be recolored):
#(utils.color-changing-cover)(rect(fill: red, stroke: blue+3pt)[test])

#{
  set rect(stroke: blue+3pt, fill: red)
  (utils.color-changing-cover)(rect[test outer scope])
}

== image slide

normal content. #pause Even more normal content.

#figure(
  image("image.png", width: 5em, height: 5em),
  caption: [A #text(red)[random] image]
)
// 


== cetz slide
normal content. #pause Even more normal content.
#import "@preview/cetz:0.5.2"
#let cetz-canvas = touying-reduce.with(cetz)
#figure(
  cetz-canvas({
    import cetz.draw: *
    rect((0, 0), (5, 5))
    (pause,)
    rect((0, 0), (1, 1))
    line((0, 0), (2.5, 2.5), name: "line")
  }),
  caption: "A cetz diagram"
)

== Tilings

#let pat = tiling(size: (30pt, 30pt))[
  #place(line(start: (0%, 0%), end: (100%, 100%)))
  #place(line(start: (0%, 100%), end: (100%, 0%)))
]
Normal content #pause
#rect(fill: pat, width: 100%, height: 60pt, stroke: 1pt)


#context rect(fill: pat, width: 100%, height: 60pt, stroke: 1pt).fill
// #context rect(fill: pat, width: 100%, height: 60pt, stroke: 1pt).stroke

== Gradients

#set text(fill: gradient.linear(red, blue))
#let rainbow(content) = {
  set text(fill: gradient.linear(..color.map.rainbow))
  box(content)
}

This is a #pause gradient on text, #pause but with a #rainbow[twist]!

// #gradient.linear(red, blue).relative()
```
requires giving it an `image.png` to compile with.


zral0kh:
well this had an issue with style rules applying colors to other elements other than text, did not check for those. should be fixed with the next one

zral0kh:
last main fix is in.

test files and documentation have to still be done. i don't have time for that anymore rn

zral0kh:
docs are done :) (@OrangeX4)

BTW since this is quite a large change for the users perhaps we want to include this in a 0.8.0?
and maybe for that we can do the speaker-note stuff as well

edit:
perhaps we can wait for typst 0.15.0 and check whether that breaks anything

tests/features/cover/ref/5.png
OrangeX4, line 1: Why there is a uncovered area beneath `Big ff text`? Is this regression or intentional?

zral0kh: the semi-transparent rect could never really work well for larger text (forgot the exact reasons but it was sth that text size can be changed by both par text size and pure text size) and it caused offsets anyways etc which is why i decided to take a totally different approach: make alpha and color changing cover the main transparent cover methods and use semi-transparent just as a fallback for those.

i could change it back but we don't need it and i really want to push for it not being used normally.
i even put a warning in when it is not used as a fallback.

perhaps it should just be taken out for text entirely? not sure

OrangeX4: If it's expected, it's okay

OrangeX4 (approved):

JoshuaLampert:
Is there anything missing here or can this be merged? Looking forward to a fix for #112 :)

zral0kh:
@OrangeX4 i created a `dev` branch which i want to use for pushing finished things to that contain breaking changes, e.g. this one or the big refactor and doc-mode or the footnotes fix

small additions and fixes can ofc go into patch versions. 

having a dev branch for those larger and breaking changes hopefully makes maintaining large branches easier

zral0kh:
we need to upgrade the tytanic version to 0.4.0 so it uses typst 0.15.0 then the tests will compile, there were some mini shift changes in layout due to typst 0.15.0 (@OrangeX4)

OrangeX4:
Yes, we can open a PR to upgrade tytanic first

JoshuaLampert:
Just to understand: What is the status of this PR? Will it be included in v0.8.0?

OrangeX4:
This is my fault. I thought that no branches were dependent on the dev branch, so I deleted the dev branch. As a result, this pull request was closed. Maybe we could re-pointing to the main branch instead of the dev branch?

OrangeX4:
@zral0kh Including #330 #364 #370, sorry for this misoperation.

zral0kh:
had not time to do anything anyway for a while. will get to it hopefully more steady from now on.

PR #378, opened 2026-06-07
Theme fix: Fix footer height for multi-line titles for `University` theme

DeepSpaceTravel:
# Motivation: 
If one has a multi-line title, it clips in the footer. This pull request is proposed to fix the problem.

# Description: 
Add dynamic adjustment to footer height from `1.5em` to `auto` to accommodate for multi-line title. 

# Testing: 
No proposed testing. Since typst layout system should work fine for `auto`.

### DIFF
--- a/themes/university.typ
+++ b/themes/university.typ
@@ -77,7 +77,7 @@
       show: block.with(width: 100%, height: auto)
       grid(
         columns: self.store.footer-columns,
-        rows: 1.5em,
+        rows: auto,
         cell(fill: self.colors.primary, utils.call-or-display(
           self,
           self.store.footer-a,


DeepSpaceTravel:
There seems to be a problem with the reference pictures, which indicates that the adjusted footer height's pixels deviate from what are in the reference. 
Is it possible to remove or skip the test?

zral0kh:
we use tytanic for the tests, so if you need to make changes there, please update the tests and commit the updated ones so we can see what changed easily.

should just be a `tt update <TEST-NAME>`

you can check with `tt run`


btw is this a problem with other themes as well? if so can you fix it there as well?

zral0kh:
do you want this to get fixed?

PR #381, opened 2026-06-10
Fix #371: cover reducer elements individually

### LINKED ISSUE #371: `fletcher-diagram` loses labeled nodes that are not in a `uncover`
Under the default context suggested by the Readme, when a `fletcher-diagram` comes after a `pause`, nodes that are in an `uncover` cannot refer to labeled nodes that are not in `uncover`. This can be relatively easily bypassed by putting the nodes under a `uncover("0-", ...)`, but a bug nonetheless.

[See full example](https://typst.app/project/rlJkhQalsgtbNJMT9l41oU)

Relevant excerpt:
```typ
= Fletcher vs pause

There is a #pause in this slide.

#let mynode(pos, name) = node(pos, "", shape: "circle", stroke: black, name: name)

#fletcher-diagram({
  mynode((0,0), "anchor1")
  uncover("0-", mynode((0,0), "anchor2"))
  
  uncover(3, node("anchor1", "x")) // Assertion failed: node `anchor1` not found.
  uncover(3, node("anchor2", "y")) // Ok.
})
```

zral0kh:
fix #371: cover elements in reducer individually instead of at the very end which messed up the ordering which can be important, e.g. for labels


OrangeX4:
Is there a test here? And will this affect the performance of the render/layout?

zral0kh:
no test added, i thought that this behaviour should have been there and we won't touch it in a bad way again anyways.

perhaps i can add the issue as a test case yes


performance wise it should not change anything, the cover of the external packages should be efficient on sets of 1 element arrays just as much as large arrays. in contrast to other stuff we do this is negligible

OrangeX4 (approved):

PR #382, opened 2026-06-10
Guard against empty output when adding last-page metadata and add regression test

### LINKED ISSUE #379: v0.7.4 breaks empty documents from parsing
Specifically in v0.7.3 if we have an empty document, with just the theme show rule it works, but not in v0.7.4;

```typst
#import themes.simple: *

#show: simple-theme.with(
  aspect-ratio: "16-9",
  config-info(
    title: [A title],
  ),
)
```

some newly introduced index is negative, which is out of bounds.

OrangeX4:
### Motivation

- Prevent a crash when compiling documents that produce no slides by avoiding indexing into an empty `output-slides` list.  
- Add a compile-only regression test to ensure empty-document compilation remains successful.  

### Description

- Change the last-page metadata logic in `src/core.typ` to only add metadata when `is-outer-call` and `output-slides.len() > 0`.  
- Improve the inline comment to document that empty documents may produce no slides and should not be indexed.  
- Add a compile-only regression test at `tests/issues/i379-empty-document/test.typ` that verifies an empty document compiles without producing visual output.  

### Testing

- Ran the automated test suite including the new regression test `tests/issues/i379-empty-document/test.typ`, and the suite passed.  
- The new compile-only tytanic test successfully compiles and prevents the previously observed crash.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a292e5dad7c832c8b50b34a44de3663)

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -1100,8 +1100,10 @@
     if slide-content != none { output-slides.push(slide-content) }
   }
 
-  //add last page metadata to last physical page.
-  if is-outer-call {
+  // Add last-page metadata to the last physical page when one exists.
+  // Empty documents can legitimately produce no slides, so avoid indexing the
+  // empty output list.
+  if is-outer-call and output-slides.len() > 0 {
     output-slides.at(-1) += [#metadata(none)#label("touying-last-page")]
   }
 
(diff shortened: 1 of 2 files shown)


zral0kh (approved):
good boi codex

Issue #327, opened 2026-03-26
Feature: config whether `pause` inside table/grid should remove following content


zral0kh:
Currently we can only cover the following cells' content of a table/grid via writing `pause`/ `waypoint`.

But for more control it should be configurable whether the following cells should just be removed, not just covered.

E.g. via a config option like `remove-covered-table-grid-cells: bool` with the current value as default: `false`.

Doing
```typst
#import "@preview/touying:0.6.3": *

#table( columns: 2, 
  [Header 1], [Header 2], pause,
  [Cell 1,1], [Cell 1,2], pause, 
  [Cell 2,1], [Cell 2,2],
)
```

might as well render as 

```typst
#import "@preview/touying:0.6.3": *

#table( columns: 2, 
  [Header 1], [Header 2],
)
```
at the first subslide, and as
```typst
#import "@preview/touying:0.6.3": *

#table( columns: 2, 
  [Header 1], [Header 2],
  [Cell 1,1], [Cell 1,2],
)
```
at the second subslide.

PS: if we ever get custom types a general option for all content whether pause should remove later content in the scope or hide that content would be nice. rn setting cover to `(body)=>[]` will just empty the cells, change the layout, but keep the cell structure
This rule would have to thus govern different content a little differently, e.g. tables. But for users it could become a simple show-set rule.

OrangeX4:
I understand the need for this, but I'm not sure how to implement it, because we might have to consider many corner cases.

1. If we put the pause inside a table cell (like `#table([Header 1 #pause], [Header 2])`), is there a way to detect exactly which cells we should remove in a nested table?
2. What is the way we can keep the blank spaces, imagine that the table is followed by something like `#meanwhile #lorem(20)`, how can we make sure that `#lorem(20)` will not move.
3. When a user might want to have the special treatment for table cells work as normal when they use different cover methods (e.g. transparent matrix masking), or when they want to remove this special treatment, how do we design the interface?

This is just my personal concern, and we can consider it.

zral0kh:
1. i mean we can track which depth we are in theoretically, so thats not an issue, all deeper levels may be removed and shallower ones not or sth like that

2. good question. we should still reserve the space of the table, just not show the stroke...: either we go ahead and while reconstructing the table set the stroke of the cells to `none`, or we wrap the table in a block with the fully expanded tables size

[...]

PR #383, opened 2026-06-10
Preserve grid/table cell metadata when covering and let absolute jumps escape outer hiding

OrangeX4:
### Motivation
- Preserve `rowspan`/`colspan` and other cell metadata when covering paused `table.cell`/`grid.cell` so layout is not lost.  
- Ensure content that can become visible earlier than its container (fn-wrappers or absolute jumps/`#meanwhile`) can escape outer hiding and manage its own visibility inside pause zones.  

### Description
- Treat absolute jumps inside cell metadata and child metadata as escapes by setting `has-fn-wrapper = true` when an absolute jump is encountered.  
- Introduce `cover-item` to apply `cover-fn` only to a cell's body and reconstruct the original `table.cell`/`grid.cell` node so cell metadata is preserved.  
- Change covered computation to map `cover-item` over items when all items are cells, and otherwise call `cover-fn(items.sum())`.  
- Add a regression test file `tests/issues/i369-i373-grid-regressions/test.typ` covering the grid/cell covering and `#meanwhile` escape cases.  

### Testing
- Added and ran the regression test `tests/issues/i369-i373-grid-regressions/test.typ`, and it passed.

### DIFF
--- a/src/core.typ
+++ b/src/core.typ
@@ -4265,9 +4265,11 @@
   // last-subslide by touying-fn-wrapper — inherit outer context so waypoints
   // placed after multi-subslide fn-wrappers fire correctly inside sub-sequences.
   let last-subslide = base-last-subslide
-  // Whether any touying-fn-wrapper was found in this parse (directly or via
-  // recursive calls).  Used by the two-pass escape hatch so that fn-wrappers
-  // inside a pause zone can handle their own visibility.
+  // Whether this parse contains content that can become visible before the
+  // surrounding repetition counter would normally reveal the whole container
+  // (fn-wrappers with their own visibility, or absolute jumps/meanwhile).
+  // Used by the two-pass escape hatch so such content can handle its own
+  // visibility inside a pause zone.
   let has-fn-wrapper = false
   // get cover function from self
   let cover = self.methods.cover.with(self: self)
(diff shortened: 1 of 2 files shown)


zral0kh:
the test ref images for tytanic are not yet added

zral0kh:
and can you maybe ask codex for whether it has a good idea for a syntax/solution for #327 ?

still can't think of anything there

edit:
perhaps we could introduce a simple `pause-cell` function and we just throw an error when it is used outside of a cell.
and during parsing we of a table/grid we store whether such a pause or a normal pause was encountered and then either cover the cells or just the content

OrangeX4:
codex said:

> For #327, I would lean toward not adding a global option that changes the meaning of normal #pause inside tables/grids. If we support this, a more explicit API like #pause-cells / #cell-pause might be clearer: normal #pause would keep the layout and cover later cells, while #pause-cells would remove later cells from the table/grid layout until they become visible. It should probably error outside direct table/grid children.
>
> That said, I’m not fully convinced this is worth adding yet. It introduces another pause variant and more rules users need to understand, especially around auto-placement, rowspan/colspan, and nested content. The extra cognitive burden may outweigh the benefit unless we see more real-world demand for this behavior.

PR #390, opened 2026-06-20
refactor: unify focus-slide with theme-level .with overrides

vincentarelbundock:
Hi all,

Thanks for your great work on `touying`!

Here's a simple PR to make focus-slide behavior and implementation more consistent across themes.

### Motivation

`focus-slide` behavior had drifted across themes and docs. There were multiple theme-local implementations and inconsistent usage patterns, which could lead to unexpected differences (e.g., defaults and background styling) even when using the same API.

### What changed

- Kept a single `focus-slide` definition in `src/slides.typ`.
- Updated themes to reuse it via `.with(...)` overrides instead of reimplementing.
- Added shared support for `background-img` and preserved configurable defaults for foreground/text styling and layout.
- Updated example/test structure so theme usage and `focus-slide` customization follow the same shared pattern.
- Moved the focus-slide regression test into `tests/features/focus-slide` with existing feature-test conventions.


vincentarelbundock:
Apologies for the earlier issues. Two should be fixed now. The third failure is something I don't understand, perhaps related to github tokens?

Let me know if I should do anything else.

src/slides.typ
```diff
@@ -3,6 +3,74 @@
...
+/// - background-img (content): Background image content (typically `image(...)` or path string).
+///
+/// - default-background (function): Returns background color when `background` and `background-color` are both `auto`/`none`.
+///
+/// - default-foreground (function): Returns text color when `foreground` is `none`.
+///
+/// - align (alignment): The alignment of the content. Defaults to `horizon + center`.
+#let focus-slide(
+  config: (:),
+  background: auto,
```
OrangeX4, line 23: Could it be similar to `page`, where only `fill` and `background` args are supported?

OrangeX4:
Sorry, you can ignore that ci issue. It’s a problem with the repository. I’ll fix it late

src/slides.typ
vincentarelbundock: Sorry, I'm not sure I understand your question. Would you prefer `focus-slide` to have fewer arguments? I feel that a lot of them are useful, but maybe we can tighten it a bit?

OrangeX4: Yes, I’m worried that too many special parameters might make it difficult for users to understand

OrangeX4:
The CI issue has been fixed. You can now rebase the branch.

src/slides.typ
vincentarelbundock: This was the union of all existing themes. I agree with you that it is too complicated, but I don't think we can be quite that minimalist. Theme owners will want to customize the default background color, image, text color, and text position. I'd maybe simplify to: `page-fill`, `text-fill`, `text-align`, `page-image`. But those words are breaking with the current API, and I'm not sure what you prefer. (Feels like the current `background` is inconsistent with Typst and not used too many other places in Touying.

vincentarelbundock: To be clear, I don't have very strong views on this, except that the function should behave consistently across themes, and that `touying` should find good nomenclature for the arguments. I'm happy to change the PR to do whatever you prefer.

PR #397, opened 2026-06-30
Add a function left-mid-right to place content on three columns

thomas-saigre:
For a theme I'm developing, I needed to create a function that split in three the content of the footer (Author, Title, and page number).

If this function can be needed in the package in the future, I gladly share it :)

*NB :* I think there is way to improvement, for instance the title may not be centered...

Thanks for the package !

### DIFF
--- a/src/components.typ
+++ b/src/components.typ
@@ -359,6 +359,20 @@
   left, none, right,
 )
 
+/// Place three content blocks at the left, in the middle and right edges of the available width using a five-column grid.
+///
+/// - left (content): The content of the left part.
+///
+/// - mid (content): The content of the middle part.
+///
+/// - right (content): The content of the right part.
+///
+/// -> content
+#let left-mid-right(left, mid, right) = grid(
+  columns: (auto, 1fr, auto, 1fr, auto),
+  left, none, mid, none, right,
+)
+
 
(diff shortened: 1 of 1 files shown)


zral0kh:
Seems useful, thank you.

If you could add the alignments, or at least make them optional or sth (then with sensible defaults), i'd be happy to merge that.

OrangeX4:
Should the “left-mid-right” function here consider centering the “mid” element? At least for me, I’d prefer that

```typ
#set page(width: 20em)

#let left-mid-right(l, m, r) = grid(
  columns: (1fr, auto, 1fr),
  l, m, align(right, r),
)

#left-mid-right[Longer left][center][right]

#left-mid-right[l][m][r]
```

image


For comparison:

```typ
#set page(width: 20em)

#let left-mid-right(left, mid, right) = grid(
  columns: (auto, 1fr, auto, 1fr, auto),
  left, none, mid, none, right,
)

#left-mid-right[Longer left][center][right]

#left-mid-right[l][m][r]
```

image

Edited: I just noticed that the issue of centering was mentioned. You can refer to that.

zral0kh:
lgtm

OrangeX4 (approved):

PR #399, opened 2026-07-03
fix: footnotes shown before pause

### LINKED ISSUE #385: Footnotes are shown even after pause
Hello,

When using a footnote after a #pause, the footnote is still shown in the previous subslide, whereas it should only appear on the subslides after the pause.

Minimum reproducible example:
```typ
#import themes.simple: *

= Title

== First Slide

Hello, Touying!

#pause

Hello, Typst!#footnote[A footnote]
```

RivinHD:
Hi,

I encountered the same issue as https://github.com/touying-typ/touying/issues/385 when writing slides and decides to quickly fix it.

Hope that helps.


zral0kh:
the fix currently does not seem to work for me. with the example from the issue the first slide still has the footnote.

perhaps you can add an issue test as well so the fix is documented properly

RivinHD:
A sorry it only fixes the issue for footnote that are provided by the bibliography-as-footnote function.
I will check if I can fix the normal footnote too.

RivinHD:
@zral0kh Ok, I fixed the normal footnote too. Can you update the new regression item. My font on my system (Linux, Fedora 44 KDE) somehow is different than what the runner has and the slides have a deviation. Can you update the new regression test, so the runner is happy: `tt update issues/i385-footnote-after-pause`

zral0kh:
Okay i checked the topic more thoroughly and it is not as easy.

First off, we cannot simply not display anything if the footnote should be hidden, this changes the layout which we want to avoid at all costs! Secondly this is not compatible with other cover methods like `color-changing-cover`.

This brings us to the problem: this seems to be an inherent limitation by typst's current `hide` implementation. We have to engineer around that and show no body and no separator for the footnote when it should be hidden. But here we face another problem: footnote.entry can only be manipulated before a page which we then cannot use to still be interoperable with user defined footnote.entry styles since we would need to capture those, ...

I first had an approach which rebuilt the footnote separator when we actually want it and normally not showed it when we also don't want to show the footnote bodies.

Now i switched back to a similar approach to before @RivinHD and decided to just keep a shadow counter and inject the markers manually into the hidden-parts when necessary.

zral0kh:
as the commit message explained the following things now apply to the functionality of the code:
- footnotes now get properly not shown with hide cover, text layout is preserved by a mockup of the super with numbering increase, which also allows the user to manipulate the numbering if they wish to
- expose the hide cover as `hiding-cover` for easy equiv, this should now be the preferred way to invoke the default hide cover. perhaps we can change the name
- config option `cover-hides-footnote` to toggle whether the active cover should show the footnote even when its covered or not, default auto can only be used with the exposed methods. when `auto` it is always false except for the `hiding-cover` function
- footnote super mockup is show-rule aware via config option `footnote-style` which should get passed a styling function. it also emits a show rule if the user solely relies on it, thus it can be used alone or to keep parity when layout things change bc of styling

- also fixed a bug introduced by the hidden heading bug fix: trail hide calls would get dropped

zral0kh:
@OrangeX4 

what do you think about the `hiding-cover` name? 

and yes this gave us two new config options, but footnotes really are nasty in typst, this was the most simple way i could think of.

Moreover this requires good documentation

OrangeX4:
I quickly looked at the code, and I think the name `hiding-cover` is fine, consistent with other cover functions.

OrangeX4 (approved):

PR #400, opened 2026-07-06
fix wrong order: hidden-parts not flushed when encountering `touying-fn-wrapper`

zral0kh:
Fixes an ordering bug during cover.

Because we did not push hidden parts when encountering an `touying-fn-wrapper` they could slip past it. Basically the same as #371 which had the same problem for reducer.


zral0kh:
@OrangeX4 just so that you notice this fix

PR #405, opened 2026-07-13
Fix 387 cover spacing issues

zral0kh:
Before we changed the whole `spacing` of the block meant to fix the list cover spacing shifts.
Now we change only the above value (based on above list continuation) and added the bottom case for bottom list continuition, e.g. when meanwhile inside a list is used.

updated tests to check those cases


zral0kh:
i think with this fix and the other stuff on main we can perhaps publish v0.7.5.

OrangeX4 (approved):

Issue #406, opened 2026-07-13
Bundle export with two Touying presentations results in wrong headings


mxmerz:
I’m trying to export two touying slideshows with a single bundle file:

```typst
#import "@preview/touying:0.7.4": *
#import themes.simple: *

#document(
  "slides-1.pdf",
  [
    #show: simple-theme
    == slide 1 A
    - this should be slideshow 1
    - but the headings are from slideshow 2!
    == slide 1 B
    == slide 1 C
  ]
)

#document(
  "slides-2.pdf",
  [
    #show: simple-theme
    == slide 2 A
    this should be slideshow 2
    == slide 2 B
    == slide 2 C
  ]
)
```

However, the headings of the second slideshow show up in both PDFs,
this is the first slide of the first slideshow:

<img width="849" height="483" alt="Image" src="https://github.com/user-attachments/assets/f9fb7148-bafe-49d0-a1ae-5b41d9cab716" />

I assume this is some introspection bug in touying, or am I holding it wrong? Is there a way to fix this? Thanks a lot in advance :)

I also [asked this as a question in the Typst forum.](https://forum.typst.app/c/questions/5)

zral0kh:
bundle export is really new and we are working on bringing touying up to par with typst 0.15.0 but it takes a while and we are working on other features as well + some maintenance. Expect it to be somewhat be fixed in 0.8.0 hopefully.

thanks for the issue though

zral0kh:
could you try importing touying once for each document?

mxmerz:
Like this?

<details>

```typst

#document(
  "slides-1.pdf",
  [
    #import "@preview/touying:0.7.4": *
    #import themes.simple: *

    #show: simple-theme
    == slide 1 A
    - this should be slideshow 1
    - but the headings are from slideshow 2!
    == slide 1 B
    == slide 1 C
  ]
)

#document(
  "slides-2.pdf",
  [
    #import "@preview/touying:0.7.4": *
    #import themes.simple: *

    #show: simple-theme
    == slide 2 A
    this should be slideshow 2
    == slide 2 B
    == slide 2 C
  ]
)
```

</details>

Results in the same bug, unfortunately.

mxmerz:
I found the location in the code that queries headings (`current-heading` in `/src/utils.typ`) and applied the solution suggested in [another Typst Forum question about querying only headings in the current document](https://forum.typst.app/t/how-to-query-headings-in-current-document/9308):

```diff
[...]

PR #409, opened 2026-07-14
fix headings in bundle export (#406)

mxmerz:
When using the experimental Typst 0.15 bundle export to create multiple Touying presentations from one
source file, the headings are wrong. (#406)

This commit fixes the bug by changing the query that computes the current heading. When multiple documents are present, it now only queries for headings within the current document.


mxmerz:
The test failure (`document is not locatable`) might be fixable by updating the test runner's tytanic version from 0.3.3 to 0.4.0.

zral0kh:
yes updating the tytanic version in the ci is planned

PR #414, opened 2026-07-31
fix: note-style citation footnotes shown before pause

SchrodingerBlume:
Continues #385.

#399 fixed a literal `#footnote` shown before `#pause`, but a citation rendered as a footnote (a CSL note style such as `chicago-notes` or `gb-7714-2015-note`) still leaked its entry onto the earlier subslides. The citation only becomes a footnote during the layout/citation pass, after the body scan that `cover-hides-footnote` relies on, so `hide()` hides the marker while the queued entry leaks through.

This handles `cite` and `ref` with the same branch structure as `footnote`: under a genuinely hiding cover, a covered citation is not rendered at all, so no footnote entry is created. It is rendered normally once revealed on a later subslide. The same leak affected packages that emit footnotes through a `show std.cite` rule, and this fixes those too.

`tests/issues/i385-cite-footnote-pause` covers both a literal `#footnote` and a note-style citation after `#pause`. `tt run` passes (68/68).


zral0kh:
so the thing is we did not test bibliographies with other styles really.
chicago-notes is quite special in the way it is automatically styled by typst and this kind of clashes with what touying does somehow. 

as i explained in #415 we need the `only` around the bib from typst 0.15.0 onwards bc otherwise it will get rendered multiple times and this causes problems with the numbering. and somehow when i add the `only` it weirdly breaks the chicago style.
but notice: chicago actually works without the `only` since it somehow manages the counting differently.

i will revert my change and accept your solution but please note that this is not how you would usually get footnotes citations in touying.

zral0kh:
otherwise i think it can be merged, it does not seem to collide with the normal bib footnotes

zral0kh:
will close this in favor of PR #428 which is just applying this to dev instead so we will have it in the next version.

thanks.

Issue #408, opened 2026-07-13
Feature Request: export .pdfpc files directly as bundled assets


zral0kh:
Since typst 0.15.0 we can now export bundles with arbitrary assets. This allows easier export of `pdfpc` files.

PR #418, opened 2026-08-06
docs: replace deprecated `typst query` with `typst eval`

FestplattenSchnitzel:
When using the mentioned command with typst version 0.15 and above, the following warning will be emitted:

warning: the `typst query` subcommand is deprecated
 = hint: use `typst eval 'query(<pdfpc-file>).first().value' --in ./example.typ` instead


zral0kh:
this is superseded soon anyway pdfpc will be available via the bundle export, see #408 

Edit: i guess it is still useful. will be added

PR #419, opened 2026-08-16
Fix broken star history chart by switching to star-history.dera.page

FaintFlower:
The star history chart in the README is currently broken because of GitHub stargazer API restrictions. This change points it to the star-history.dera.page mirror, which serves both the frontend and the /svg chart from a single domain using an alternative data source that requires no API token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,10 +381,10 @@ Thanks to...
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


OrangeX4:
Thanks

PR #420, opened 2026-08-19
Theme fix: Don't show linebreak in footer with university and stargazer theme.

marisbaier:
This pull-request aims at improving the university-theme and the stargazer-theme. In these two themes, the title of the presentation is being displayed in the footer. In case you want a specific linebreak in the title-slide, at the moment, this produces weird behaviour in the footer, see screenshot below:
university_before
With small change in this pull-request this is fixed:
image

### DIFF
--- a/themes/stargazer.typ
+++ b/themes/stargazer.typ
@@ -430,6 +430,7 @@
   let footer(self) = {
     set text(size: .5em)
     set std.align(center + bottom)
+    show linebreak: none
     grid(
       rows: (auto, auto),
       utils.call-or-display(self, self.store.footer),
--- a/themes/university.typ
+++ b/themes/university.typ
@@ -68,6 +68,7 @@
   let footer(self) = {
     set std.align(center + bottom)
     set text(size: .4em)
+    show linebreak: none
     {
       let cell(..args, it) = components.cell(
         ..args,


zral0kh (approved):
looks good

zral0kh:
thank you, do you think there should be a dash or so instead?

marisbaier:
> thank you, do you think there should be a dash or so instead?

Thank you for the input! I haven't thought of that! Although what you're saying would be a good idea for a title like 

"Early against Late: \ A contrast on dark energy in the light of DESI DR2",

it would not be a good idea for linebreaks that are purely cosmetic and have no meaning (and we don't know which ones are). For example, I initially encountered this problem with the title of my thesis that was just too long and thats why I needed linebreaks but at specific positions.

zral0kh:
actually can you share your example. i just had another look at how university does its footer and it should just display the `info.title` or `info.short-title` by default if `footer-b` is set to `auto`.

this means you can likely achieve the same by just changing what you pass into those or what you pass into the theme.

marisbaier:
> actually can you share your example. i just had another look at how university does its footer and it should just display the `info.title` or `info.short-title` by default if `footer-b` is set to `auto`.
> 
> this means you can likely achieve the same by just changing what you pass into those or what you pass into the theme.

My use case boils down to this:

```
#import themes.university: *

#show: university-theme.with(
  config-info(
    title: [Title \ more title],
  ),
)

#title-slide()

A Slide
```
which compiles to this:
image
image

zral0kh:
sorry for the long delay in answering:

in this case you can just set the footer directly for the theme or write a short-title instead. 
i am not a fan of fixing this use case when other people might depend on it (unlikely tho) while there is a good work-around.

we can have this sit around for a while and wait if other people want this to be fixed too.

PR #423, opened 2026-08-25
fix: honor focus slide config before defaults

### LINKED ISSUE #421: University theme: Passing config to focus-slide has no effect
Currently passing a config to the university theme focus slide has no effect.
```
#focus-slide(config:config-colors(primary: white, neutral-lightest: black))[
  Thank you for your attention!
  
  colors.primary = #touying-get-config("colors.primary")

  colors.neutral-lightest = #touying-get-config("colors.neutral-lightest")
]
```
produces

Image

While expected would be something like

Image

This is due to the config passed into `focus-slide` is never really used. https://github.com/touying-typ/touying/blob/3d5332b0d992bee8581df4c88e6e32bf7d23c143/themes/university.typ#L290-L294 the `utils.merge-dicts` of config, doesn't include the passed one as it does for example in `#title-slide`  https://github.com/touying-typ/touying/blob/3d5332b0d992bee8581df4c88e6e32bf7d23c143/themes/university.typ#L142-L146

GeronimoCastano:
Fixes #421

The focus-slide functions calculated their default background and text values from `self` before applying the per-slide `config`. As a result, `config-colors(...)` arrived too late to affect those defaults.

This applies the focus-slide config first in all six built-in themes. The existing explicit `foreground` parameter in the simple theme remains available for callers who want to override its text color directly.

### Tests

- Added a regression test for the University theme reproduction.
- `tt run` with Typst 0.15.0 (`3ae52774`): 68/68 passed.
- `typstyle --check themes tests` passed.
- `git diff --check` passed.

### Rendered output

Before: the requested white background and black text are ignored.

Before

After: the focus slide uses the supplied colors.

After


zral0kh:
thanks

PR #426, opened 2026-09-09
Fix bibliography configuration using dictionary

teobiton:
I have been experimenting with custom numbering for my bibliography. However, using a dictionary in config-common, as explained [here](https://touying-typ.github.io/docs/reference/configs/config-common#show-bibliography-as-footnote) resulted in a compile error. These modifications in `slides.typ` fix the issue. I am new to typst so maybe I am unaware of a better way to do it using `..args` or `.with()`.

Also, I did not find an efficient way to reflect this in `tests/features/bibliography`, I am open to suggestions regarding this. It seems that using `#show: touying-set-config.with(config-common()` with bibliography has no effect on the setting.

### DIFF
--- a/src/slides.typ
+++ b/src/slides.typ
@@ -100,12 +100,13 @@
     if self.at("show-bibliography-as-footnote", default: none) != none {
       let args = self.at("show-bibliography-as-footnote", default: none)
       if type(args) == dictionary {
-        let bibliography = args.at("bibliography")
-        args.remove("bibliography")
-        magic.show-bibliography-as-footnote.with(
-          ..args,
+        let bibliography = args.at("bibliography", default: none)
+        let numbering = args.at("numbering", default: "[1]")
+        magic.bibliography-as-footnote(
           bibliography,
           body,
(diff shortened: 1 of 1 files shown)


zral0kh:
yup that was broken, i also recently saw this, thanks for pushing on it.

this will change completely anyways, we are going back to using the normal bibliography in the next version. this means the bibliography `style` will be responsible for handling the numbering again.
sorry to disappoint you there. you can work with the dev branch if you want, the bibliography change is already in there.

a quick chat with claude gave the following way to change the text markers:
```typst
#let seen = state("cite-keys", ())
#show cite: it => {
  if it.form != "normal" {
    it
  } else {
    seen.update(k => if it.key in k { k } else { k + (it.key,) })
    cite(it.key, form: none)  // registers the entry, does not show anything
    context {
      let n = seen.get().position(k => k == it.key) + 1
      [(#n)]
    }
  }
}
```

teobiton:
I see. Thanks for your answer and for the suggestion.

Closing this then.

PR #427, opened 2026-09-11
Hide unoutlined sections from mini-slides

teobiton:
Sections and slides under unoutlined headers get hidden from the mini-slides bar, and no header is highlighted on these slides.

_Note: I have been working the `dev` branch but if changes to it are not welcomed, this can be discussed after the 0.8.0 release._

### DIFF
--- a/src/components.typ
+++ b/src/components.typ
@@ -814,10 +814,12 @@
   inline: false,
 ) = (
   context {
+    // Keep every heading, even hidden or unoutlined ones: they are needed as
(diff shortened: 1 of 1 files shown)


zral0kh:
hi thx for the commit. 
no it is very welcome thx.

zral0kh:
one change: we should not abandon the whole tree but instead only the exact heading that is unoutlined since that is what normal typst does in its outline as well.

i know this does not make that much sense but we should keep parity.

PR #431, opened 2026-09-17
fix: reducers did not announce inner min reps and fn-wrappers

zral0kh:
See the new test for the cases.

was just an oversight i did not think of before...


zral0kh:
@OrangeX4 merge it whenever you want

